PR #10, opened 2014-11-01
Feature dry run

OliverTED:
Hello,

-------------------
This patch adds an option --dry-run to the run command.

In this mode changes to the database are prevented by a thin
layer around the the database object. Task action and teardown
execution is skipped.
-------------------

I did not test it thoroughly, its my first contribution :-). I did not add tests for the new code. Do you think this can be pulled ? Or what do you require for it being pulled ?

Thanks for your efforts,
Oliver


OliverTED:
Hello Eduardo,


I got a request regarding dry-run. My question is the following:

* Do you want to include the feature in doit
* What precisely are your requirements?
* I have few experience with testing and would require your help there.

My recent project using doit is finished right now. I would just 
continue to work if this is in your interest and you plan to include it.


Best,
Oliver

schettino72:
Hi Oliver, 

This feature will be a nice addition to doit. So yes, I would like to see it included in doit.

I dont have exact requirements, who implements it can more or less decide how it should be (as long as it doesnt mess up with the existing code). 

Sure I can help and guide the implementation (as I already contributed to this feature). 
But if you leave it unfinished dont expect me to pick it up and do all the work...

Cool you finished your project. what is it? is it open-source? do you think you can contribute a success story?

OliverTED:
Hello Eduardo, 


Well it is implemented and mostly working. Tom did a few tests and
errors appeared and I guess there will be a few more when it is used and
I will fix them as they appear.

Still there is this database layer that you were not so happy with (and
I understand why). Yet to decide if things are to be build one needs to
store what has been built already. While some dbs buffer changes and
could simply be kept from writing back to disc, other do not and require
such a layer. So why not use it for all databases. Trying to keep
influences on other code small the layer seems convenient. 

BTW Gnu make utilizes -j to set number of concurrent processes (right
now -n in pydoit) and -n for dry run. Possibly one could adjust pydoit's
to match gnu make. 

The project I finished was just some specialized University stuff, not
particularly interesting for the public. 

This is the first time I contribute, can you tell me what you require,
e.g.
* remove the layer from backends other than sql ? Would it make it
better ?
* write tests cases ?

I would be happy to finish it and never meant you to do the final work.
Just I do not have a lot of time.


Best, Oliver 


On January 7, 2015 1:51:29 PM GMT+01:00, eduardo naufel schettino
<notifications@github.com> wrote:
        Hi Oliver, 
        
        This feature will be a nice addition to doit. So yes, I would
        like to see it included in doit.
        
        I dont have exact requirements, who implements it can more or
        less decide how it should be (as long as it doesnt mess up with
        the existing code). 
        
        Sure I can help and guide the implementation (as I already
        contributed to this feature). 
        But if you leave it unfinished dont expect me to pick it up and
        do all the work...
        
        Cool you finished your project. what is it? is it open-source?
        do you think you can contribute a success story?

schettino72:
I think the DB layer can be removed altogether... I did it here (https://github.com/pydoit/doit/commit/b2d88fc10a02a508989f8f28625ec57548d05698) and it seems to work. If DB layer is required you need to show me an example and convince me :)

Having the same options as GNU Make was never a goal for me. i dont remember where i got the idea of `-n`... what `-j` and `-n` stand for on Make? pick whatever letter makes some sense for dry run but I dont think it is worth changing `-n` already taken for parallel.

Yes. It also need unit-test but it will be a waste of time if you write and the code changes too much (like removing the DB layer).

Please check my branch (show me an example where it fails and your DB layer doesnt). And then ping me when it is ready for review.

schettino72:
oh. I see I chose `-n` for the the `dry-run` option on the clean command... it better be the same as for the `run` command.

OliverTED:
Yes, I guess you are right. As I wrote, difficult to get a good
understanding - I assumed sql directly writes back.

Clearly if the databases do not write back intermediate changes - then
removing DryRunDB is the better solution.

JsonDB and SqliteDB do not commit intermediate changes. DbmDB appears to
commit changes when 'remove' or 'remove_all' is called.

As far as I see they are only called in 'doit forget'. So it should not
pose a problem in 'doit run --dry-run'. Yet it appears unclean.

Another thing I am unsure of is your added lines:
#    def _set(self, task_id, dependency, value):
#        if not self.dry_run:
#            self.backend.set(task_id, dependency, value)
Are they required? Is there a backend still writing back in dry run
mode?


What do you think ?

Oliver

schettino72:
> Another thing I am unsure of is your added lines:
```
     def _set(self, task_id, dependency, value):
         if not self.dry_run:
             self.backend.set(task_id, dependency, value)
```
> Are they required? Is there a backend still writing back in dry run mode?

All current implementations caches the changes as a performance optimization.
But this makes them unsuitable for usage by more than one instances of doit running at the same
time. I would like to leave the possibility of a set modifying the storage open for future implementations.

### Other comments

  * why did you change the signature of `Runner.__init__` ? If not necessary please revert.
  * the codes need some cleanup after all the changes in logic... and could you rebase it to master?
  * can you document the behaviour of the implementation? makes it easier to review.
     I can see you ignore missing file_dep, I don't like it... it should be ignored only if the missing file is 
     a target from another file. See the point below

 * what happens if you have:

                |   input   | output
-------------|-----------|--------------
task_a    |   foo.x    |   foo.y
task_b    |   foo.y    |   foo.z

If foo.x changed? I think task_a and task_b should be considered that they will change... 
does `make` handle this?

 * better create a sample dodo.py where we can test some scenarios.

OliverTED:
Hello,


finally I managed to spend some time. How do manage doit and a job? By 
the way, sorry for delays. New version is on my master branch of 
OliverTED/doit.

>
>       Other comments
>
>   * why did you change the signature of |Runner.__init__| ? If not
>     necessary please revert.
Is reverted.


>   * the codes need some cleanup after all the changes in logic... and
>     could you rebase it to master?
Done.

>   * can you document the behaviour of the implementation? makes it
>     easier to review.
I tried to document my thought as well as possible in the code.


>     I can see you ignore missing file_dep, I don't like it... it should
>     be ignored only if the missing file is
>     a target from another file. See the point below
>
>       o
>
>         what happens if you have:
>
>         	input 	output
>         task_a 	foo.x 	foo.y
>         task_b 	foo.y 	foo.z
>
> If foo.x changed? I think task_a and task_b should be considered that
> they will change...
> does |make| handle this?
In the best case this clears up with the comments.

>   * better create a sample dodo.py where we can test some scenarios.

I created a directory 'mytests' - clearly not for inclusion in the main 
branch but for testing. Note that Tom Varga helped a lot with testing!




Best,
Oliver

schettino72:
Hi Oliver.

First, thanks for working on this... 

>> If foo.x changed? I think task_a and task_b should be considered that
 they will change...
 does |make| handle this?

> In the best case this clears up with the comments.

I have no idea what you mean. What exactly is being implemented needs to be clear...

I wrote down some use-cases here:
https://gist.github.com/schettino72/9a798a73189567d136cd

As you can see there, there are 2 instances where it has some unexpected behaviour to me.
And in one case also differs from `make`. 
Please add any other use-case that is important to you... 

So before we even look at the implementation lets agree on what needs to be implemented.
We should have set this clearer before...

OliverTED:
Hello Schettino,

thanks for your nice feedback. If it helps, I enjoy spending a bit time on this, just that I only have time occasionally. Your setup use cases help a lot, as this gives me a guide how you prefer them handled.

Generally I thought that any rule that could be called should be printed on dry run. (Do you agree on this principle?). Then automatically you have the equivalence of --continue and --dry-run (or am I mistaken). Because --continue should not build anything that eventually changes by a successful run of failed rules.

I implemented missing dependencies more relaxed to get out of the way of the user. But on a second thought I agree that early error reporting should be preferred. So a second principle could be strict error checking, i.e. report malformed build rules as soon as they can be detected.

I will go through those reported use cases and update the behavior.


Best / Oliver

OliverTED:
Early error reporting might be fixed. Your last test case does only run t1 at my installation, independent whether called with/without --dry-run. But I will take some more time those days to evaluate. I did not really understand task_dependencies - as I did not use/require them yet.

schettino72:
> Generally I thought that any rule that could be called should be printed on dry run. (Do you agree on this principle?). Then automatically you have the equivalence of --continue and --dry-run (or am I mistaken). Because --continue should not build anything that eventually changes by a successful run of failed rules.

You are using `make` terminology :) it does not exactly map to `doit` concepts so I am bit confused by what you mean.  I would say:

`--continue` should not ~~build~~ try to execute anything that ~~eventually changes by a successful run of~~ depends on failed tasks ~~rules~~.

For my last [task_dep example](https://gist.github.com/schettino72/9a798a73189567d136cd#task_dep--uptodate) I included the sequence of commands to test it. The expected behaviour on the table refers to last invocation of `doit`. It is a poor example because I set `uptodate:True` it should be something that would fail if `t1` is executed again... It would be better example if it had a `result_dep` and `t1` would change it is returned value on every execution.

If we agree with the expected behaviour we can move to the implementation.

### doit internals

`doit` execution basically has three components:
  - control.py controls the execution order of the tasks
  - dependency.py checks if a task is up-to-date or not
  - runner.py get tasks from control.py, check if need to be executed using dependency.py and finally executes task's actions.

dependency.py has absolutely NO knowledge about relation between tasks or the order they should be executed. It always deals only with a single task.

control.py during its initialization will check for `file_dep` of all tasks that are a `target` from another one. For those it will implicitly create a `task_dep`. After this point control.py doesn't care at all about `file_dep` or `targets`.

`doit` is general purpose tool and not strictly file-based as `make`. But since files based operations is so common doit has some built-in support for it by translating these file relations between tasks into the more general principle of `task_dep` that is used to build the task tree.

### dry-run implementation

depenendency.py: I am now wonder if `dependency.py` should not be modified at all. At most it should contain only the modifications up to here (https://github.com/pydoit/doit/compare/feature-dry-run). The other option would be to modify `runner.py` to not call any operations that modify the DB.

The problem of your implementation is that you make changes on dependency.py so it will only be able to handle file dependencies. Handling things on control.py will allow you handle the more general case of `task_dep`. 

Note that since `doit` already has support --continue some machinery on how to deal with propagation of successful/unsuccessful status of task is already present. 

So for dry-run I guess that we could say that if task has the status `run` you could imply that all tasks that depends on it will also need to be `run` without even checking if it is up-to-date by checking its `file_dep`s. This way `file_dep` that doesnt exist but are supposed to be created by other tasks will not raise error because the up-to-date check (`[runner]self.dep_manager.get_status()`) should not event be called ...

ps:

* all this things were not clear to me also, just now after looking at your implementation and thinking about use-cases it became clear. sorry it took some time of you going in the wrong direction for me to realize that.

* no worries about working on it just occasionally, take your time.

schettino72:
closing this since it is inactive. I created an issue to keep track for future attempts.

PR #15, opened 2015-01-07
listen for MOVED_TO events in doit auto

FlorianLudwig:
If a file is replaced it does change too.

For the OS X code path this change should be applied as well.  From a quick search fsevents.IN_MOVED_TO seems to exist but I have not verified it.

Btw. actually the safe way to "save" a file is to create a temp file and then move it over the original. So no other concurrently running process could read only half of it or in case of power failure the file would not corrupt.

### DIFF
--- a/doit/filewatch.py
+++ b/doit/filewatch.py
@@ -97,7 +97,7 @@ class FileModifyWatcher(object):
         event_handler = EventHandler()
         notifier = pyinotify.Notifier(watch_manager, event_handler)
 
-        mask = pyinotify.IN_CLOSE_WRITE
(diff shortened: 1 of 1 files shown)


schettino72:
Did you test this change? You mention OS X but make changes in the code that is only executed on linux!

Please check this: http://schettino72.wordpress.com/2010/03/07/inotify-text-editors-emacs-vim/
`IN_CLOSE_WRITE` seems to handle all cases on linux. 

On OS X it is a different story... if you can describe a way to reproduce the problem would be great. Also you play on printing all received events (as I did in the blog) should help.

schettino72:
thanks!

PR #21, opened 2015-02-02
Documentation cleanup

noisecapella:
Moved types into @type fields so that epydoc processes them correctly

### DIFF
--- a/doit/action.py
+++ b/doit/action.py
@@ -89,18 +89,23 @@ class CmdAction(BaseAction):
     """
     Command line action. Spawns a new process.
 
-    @ivar action(str,list,callable): subprocess command string or string list,
+    @ivar action: subprocess command string or string list,
          see subprocess.Popen first argument.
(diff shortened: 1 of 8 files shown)


schettino72:
Thanks for the patch. I didnt expect anyone to care about this generated API :)

* unfortunately epydoc looks like a dead project. No python 3 support and no updates in the last 5 years.
* it seems sphinx killed it... although I still think there is space for a pure API tool like epydoc.
* Actually I am aware that I use the wrong syntax for types, but I do that on purpose because I actually never look at the generated docs myself and I think using 2 lines is too verbose.
* sphinx actually support passing the `type` in the same line as long as it is only a single word. http://sphinx-doc.org/domains.html#info-field-lists

So I don't see much point in trying to fix things for epydoc. I would rather move to sphinx or something else that is actively being developed. If not sphinx should be something that uses the ReST format same as sphinx so that the API can be used by both tools.

Issue #22, opened 2015-02-05
Option to not use md5 for file dependencies.


saimn:
Hi,
First, thanks for doit, it is a great tool !
I'm using it currently to process a lot of huge files, and the bottleneck is the computation of the md5sums for the file dependencies. From what I have seen in the code it is not possible currently to have this. Even if I use `task_dep` and `uptodate` with `check_timestamp_unchanged` instead of `file_dep`, I still need to specify `targets`, hence md5sum is computed on the target files. 
So, unless I missed another way to achieve this, what I would like is to use `file_dep` and `target`, as it works well and is very practical with subtasks, but with only the timestamp check. 
Maybe adding an option in `DOIT_CONFIG` to deactivate the use of md5sums could be a solution ?

schettino72:
It would be nice to add a way plug a different implementation for check if a file changed (not only being able to disable MD5).

I see this could be done in 3 different levels.
- globally on `DOIT_CONFIG`
- on task level with a new attribute
- for a each `file_dep`

Did you actually measure it? I would like to see some benchmarks (or your code is open source?)

md5 is never computed for targets, where you get this from?

Another problem is that I think doit is calculating the md5 of same file more than once per run.

saimn:
> It would be nice to add a way plug a different implementation for check if a file changed (not only being able to disable MD5).

Yes, indeed !

> Did you actually measure it? I would like to see some benchmarks (or your code is open source?)

No the code is not open source. A little test with a 'echo' action and a 8.5Gb file takes roughly ~40sec.

> md5 is never computed for targets, where you get this from?

My bad, I looked at the code a few days ago and I didn't remember correctly. So yes, it is computed only for `file_dep`.

Issue #7, opened 2014-10-01
unstable test on travis


schettino72:
There are some tests that sometimes fails... I works consistently on my machine but fails on travis.
It seems there is no way to debug on travis.

PR #23, opened 2015-02-06
Allow to customize how file_dep are checked.

saimn:
Following #22, I tried to do something, but I don't understand clearly how the config / params are handled. Comments welcome ;)

Add a `modified_checkers` option which allows to choose how file_dep are checked
to know if they are modified. `modified_checkers` is a dict with boolean values
for each checker:
```
DOIT_CONFIG = {
    'modified_checkers' : {
        'timestamp': True,
        'size': True,
        'md5': True
    }
}
```

### DIFF
--- a/doc/cmd_run.rst
+++ b/doc/cmd_run.rst
@@ -44,7 +44,8 @@ This flag is valid for all sub-commands.
     $ doit -f release.py
 
 
-*doit* can seek for the ``dodo.py`` file on parent folders if the the option ``--seek-file`` is specified.
+*doit* can seek for the ``dodo.py`` file on parent folders if the option
+``--seek-file`` is specified.
 
 
 as an executable file
@@ -277,7 +278,7 @@ check_file_uptodate
 `doit` provides different options to check if files are up to date.
 Use the option ``--check_file_uptodate`` to choose:
 
- * default: use the timestamp, size and md5sum of the file.
+ * md5: use the md5sum of the file.
  * timestamp: use only the timestamp and size.
 
 .. code-block:: console
@@ -291,18 +292,18 @@ custom check_file_uptodate
 It is possible to define your own custom up to date checker. Check the code on
 `doit/dependency.py
 <https://github.com/pydoit/doit/blob/master/doit/dependency.py>`_ ... It is
-easy to get started by sub-classing the default reporter as shown below. The
-custom reporter must be configured using DOIT_CONFIG dict.
+easy to get started by sub-classing ``FileChangedChecker`` as shown below. The
+custom checker must be configured using DOIT_CONFIG dict.
 
 .. code-block:: python
 
-    from doit.dependency import BaseChecker
+    from doit.dependency import FileChangedChecker
 
-    class MyChecker(BaseChecker):
+    class MyChecker(FileChangedChecker):
         """With this checker, files are always out of date."""
         def check_modified(self, file_path, state):
             return True
-        def save_state(self, task, dep):
+        def get_state(self, task, dep, current_state):
             pass
 
     DOIT_CONFIG = {'check_file_uptodate': MyChecker}
--- a/doc/cmd_run.rst
+++ b/doc/cmd_run.rst
@@ -303,7 +303,7 @@ custom checker must be configured using DOIT_CONFIG dict.
         """With this checker, files are always out of date."""
         def check_modified(self, file_path, state):
             return True
-        def get_state(self, task, dep, current_state):
+        def get_state(self, dep, current_state):
             pass
 
     DOIT_CONFIG = {'check_file_uptodate': MyChecker}
--- a/doc/cmd_run.rst
+++ b/doc/cmd_run.rst
@@ -271,6 +271,42 @@ from a task while it is being executed,
 this is controlled by the ``verbosity`` param.
 
 
+check_file_uptodate
+-------------------
+
+`doit` provides different options to check if files are up to date.
+Use the option ``--check_file_uptodate`` to choose:
+
+ * default: use the timestamp, size and md5sum of the file.
+ * timestamp: use only the timestamp and size.
+
+.. code-block:: console
+
+    $ doit --reporter timestamp
+
+
+custom check_file_uptodate
+--------------------------
+
+It is possible to define your own custom up to date checker. Check the code on
+`doit/dependency.py
+<https://github.com/pydoit/doit/blob/master/doit/dependency.py>`_ ... It is
+easy to get started by sub-classing the default reporter as shown below. The
+custom reporter must be configured using DOIT_CONFIG dict.
+
+.. code-block:: python
+
+    from doit.dependency import BaseChecker
+
+    class MyChecker(BaseChecker):
+        """With this checker, files are always out of date."""
+        def check_modified(self, file_path, state):
+            return True
+        def save_state(self, task, dep):
+            pass
+
+    DOIT_CONFIG = {'check_file_uptodate': MyChecker}
+
 
 output-file
 ------------
--- a/doc/cmd_run.rst
+++ b/doc/cmd_run.rst
@@ -275,15 +275,15 @@ this is controlled by the ``verbosity`` param.
 check_file_uptodate
 -------------------
 
-`doit` provides different options to check if files are up to date.
-Use the option ``--check_file_uptodate`` to choose:
+`doit` provides different options to check if dependency files are up to date
+(see :ref:`file-dep`).  Use the option ``--check_file_uptodate`` to choose:
 
- * md5: use the md5sum of the file.
- * timestamp: use only the timestamp and size.
+ * md5: use the md5sum.
+ * timestamp: use the timestamp.
 
 .. code-block:: console
 
-    $ doit --reporter timestamp
+    $ doit --check_file_uptodate timestamp
 
 
 custom check_file_uptodate
--- a/doit/cmd_base.py
+++ b/doit/cmd_base.py
@@ -271,7 +271,6 @@ class DoitCmdBase(Command):
         self.dep_class = backend_map.get(params['backend'])
         params['pos_args'] = args # hack
         params['continue_'] = params.get('continue') # hack
-        params['modified_checkers'] = self.config.get('modified_checkers') # hack
 
         # magic - create dict based on signature of _execute method
         args_name = inspect.getargspec(self._execute)[0]
--- a/doit/cmd_base.py
+++ b/doit/cmd_base.py
@@ -260,8 +260,8 @@ class DoitCmdBase(Command):
         if minversion:
             if version_tuple(minversion) > version_tuple(version.VERSION):
                 msg = ('Please update doit. '
-                'Minimum version required is {required}. '
-                'You are using {actual}. ')
+                       'Minimum version required is {required}. '
+                       'You are using {actual}. ')
                 raise InvalidDodoFile(msg.format(required=minversion,
                                                  actual=version.VERSION))
 
@@ -271,6 +271,7 @@ class DoitCmdBase(Command):
         self.dep_class = backend_map.get(params['backend'])
         params['pos_args'] = args # hack
         params['continue_'] = params.get('continue') # hack
+        params['modified_checkers'] = self.config.get('modified_checkers') # hack
 
         # magic - create dict based on signature of _execute method
         args_name = inspect.getargspec(self._execute)[0]
--- a/doit/cmd_run.py
+++ b/doit/cmd_run.py
@@ -100,11 +100,11 @@ opt_check_file_uptodate = {
     'short': '',
     'long': 'check_file_uptodate',
     'type': str,
-    'default': 'default',
+    'default': 'md5',
     'help': """\
 Choose how to check if files have been modified. Available:
-'default': use timestamp, size and md5sum
-'timestamp': use only timestamp and size
+'md5': use the md5sum
+'timestamp': use the timestamp
 """
 }
 
@@ -148,7 +148,7 @@ class Run(DoitCmdBase):
     def _execute(self, outfile,
                  verbosity=None, always=False, continue_=False,
                  reporter='default', num_process=0, par_type='process',
-                 single=False, check_file_uptodate='default'):
+                 single=False, check_file_uptodate='md5'):
         """
         @param reporter:
                (str) one of provided reporters or ...
--- a/doit/cmd_run.py
+++ b/doit/cmd_run.py
@@ -133,7 +133,7 @@ class Run(DoitCmdBase):
     def _execute(self, outfile,
                  verbosity=None, always=False, continue_=False,
                  reporter='default', num_process=0, par_type='process',
-                 single=False):
+                 single=False, modified_checkers=None):
         """
         @param reporter:
                (str) one of provided reporters or ...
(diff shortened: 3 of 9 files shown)


doit/cmd_base.py
```diff
@@ -271,6 +271,7 @@ def execute(self, params, args):
         self.dep_class = backend_map.get(params['backend'])
         params['pos_args'] = args # hack
         params['continue_'] = params.get('continue') # hack
+        params['modified_checkers'] = self.config.get('modified_checkers') # hack
```
saimn: This part is quite obscure for me, so I ended up with this for the moment.

saimn:
The idea is that the dict could contain custom implementation for the different checkers (not implemented yet), but I'm not sure if this would work correctly with the multiprocessing runner.

Also the tests passed locally, so I am not sure if the TestFileWatcher failure on Travis is related to my changes ?

schettino72:
Thanks for the patch. I took just a quick look...

> The idea is that the dict could contain custom implementation for the different checkers (not implemented yet),

I dont think using a dict like that is a good idea. Maybe the combination of values doesnt make sense...
I would prefer if it was done the same as `reporters` are done. That is, built-in supported can be configured using its name "string" or you pass an object. I guess this object will need at least 2 methods as a minimum interface. One method for checking and another one for saving.

> Also the tests passed locally, so I am not sure if the TestFileWatcher failure on Travis is related to my changes ?

#7

doit/cmd_base.py
schettino72: sorry. i need to write some better comments here.

saimn:
> I dont think using a dict like that is a good idea. Maybe the combination of values doesnt make sense...
I would prefer if it was done the same as reporters are done. That is, built-in supported can be configured using its name "string" or you pass an object.

I chose this approach because currently the 3 options are used (timestamp, size, md5), but it seems possible to use any combination of the 3. But it was just an idea, I'm open to any solution, I just want to disable the md5 ;-).
I looked at how it is done for reporters, but I don't understand how the same can de done here. Could you give an example of what the DOIT_CONFIG should look like ?
Also I didn't understand how the DOIT_CONFIG / task params are handled in the code, so I would appreciate some clues on where to declare the default value, how to overload DOIT_CONFIG with task params, ...

schettino72:
> I chose this approach because currently the 3 options are used (timestamp, size, md5), but it seems possible to use any combination of the 3. But it was just an idea, I'm open to any solution, I just want to disable the md5 ;-).

Not all combinations make sense. Checking the size is just an optimization. Currently the timestamp check just check if the time is exactly the same, not like `make` where it compares the timestamp of the targets against the dependencies...

> I looked at how it is done for reporters, but I don't understand how the same can de done here. Could you give an example of what the DOIT_CONFIG should look like ?

doit config should like this: http://pydoit.org/cmd_run.html#custom-reporter you can specify a string or a class (before i said object, class or object are both ok maybe class is better for consistency).
The code looks like this: https://github.com/pydoit/doit/blob/master/doit/cmd_run.py#L159

> Also I didn't understand how the DOIT_CONFIG / task params are handled in the code, so I would appreciate some clues on where to declare the default value, how to overload DOIT_CONFIG with task params, ...

You can implement just what you need... just look at the code. `verbosity` can be defined both at the DOIT_CONFIG and at task level. There is some magic on how the parameters are passed to the `_execute` method (thats why there are some lines with `# hack` marked, those are parameters I add to match the name on `_execute()` definition).

schettino72:
To make it clear... I would like to see a class on dependency.py with two methods: `check_modified()` and `save_state()` where the class with the default implementation is used by default. 

Add another class with the same interface with an implementation that does not use MD5. This second implementation could be selected by passing a string to `DOIT_CONFIG['check_file_uptodate']="timestamp"`. 

This config could also take a python class (instead of a string) with the same interface.

To support this at the task level is a bit more complicated and I am not sure all the steps that would be required to accomplish that. we can focus on setting at the global level first. if you also want on task level we look at it later...

saimn:
Thanks for all the explanations, it is much clearer now :).

saimn:
@schettino72 : It should be better now, if you want to take a look.

doit/dependency.py
```diff
@@ -361,6 +338,96 @@ def remove_all(self):
...
+    def __init__(self, backend):
+        self.backend = backend
+        self._set = self.backend.set
+        self._get = self.backend.get
+
+    def check_modified(self, file_path, state):
+        """Check if file in file_path is modified from previous "state".
+
+        @param file_path (string): file path
+        @param state (tuple), timestamp, size, md5
```
schettino72: The state (timestamp, size, md5) is an implementation detail of the current implementation...

We happens if a new checker needs to save different information?
What if the user runs `doit` once with one checker, and then again with a different checker?

doit/dependency.py
```diff
@@ -361,6 +338,96 @@ def remove_all(self):
...
+        """Save info after a task is successfuly executed"""
+        pass
+
+
+class DefaultChecker(BaseChecker):
+    """Default checker. Use the timestamp, size and md5sum."""
+
+    def check_modified(self, file_path, state):
+        try:
+            file_stat = os.stat(file_path)
```
schettino72: Calling os.stat() many times will cause a performance degradation for users who have thousands of small files...

Thats why I had the trouble to save os.stat() result and re-use it. Can you avoid it?

doit/dependency.py
```diff
@@ -361,6 +338,96 @@ def remove_all(self):
...
+            return True
+
+        timestamp, size, _ = state
+
+        # 1 - if timestamp is not modified file is the same
+        if file_stat.st_mtime == timestamp:
+            return False
+
+        # 2 - if size is different file is modified
+        return file_stat.st_size != size
```
schettino72: this size check alone doesnt make sense. what you are expecting from this?

doit/dependency.py
```diff
@@ -361,6 +338,96 @@ def remove_all(self):
...
+
+    def save_state(self, task, dep):
+        timestamp = os.path.getmtime(dep)
+        size = os.path.getsize(dep)
+        self._set(task.name, dep, (timestamp, size, None))
+
+
+# name of checkers class available
+CHECKERS = {'default': DefaultChecker,
```
schettino72: I would rather call it `md5` or `md5sum` than `default`.

doit/dependency.py
```diff
@@ -361,6 +338,96 @@ def remove_all(self):
         self._conn.execute('delete from doit')
 
 
+class BaseChecker(object):
+    """Base Checker that must be inherited."""
+
+    def __init__(self, backend):
```
schettino72: I think it is not necessary to take the `backend` as a parameter.

on `save_state` pass the current state, and return the state to be saved...
`save_state()` was probably not a good name.

doit/dependency.py
```diff
@@ -438,7 +497,7 @@ def get_value(self, task_id, key_name):
         """
         if not self._in(task_id):
             # FIXME do not use generic exception
-            raise Exception("taskid '%s' has no computed value!" % task_id)
+            raise Exception("taskid '%s' has no computed v, DefaultCheckeralue!" % task_id)
```
schettino72: ???

doit/dependency.py
schettino72: ok. ignore my comment. I can see you call it just once as before.

doit/dependency.py
```diff
@@ -361,6 +338,96 @@ def remove_all(self):
         self._conn.execute('delete from doit')
 
 
+class BaseChecker(object):
```
schettino72: The name `checker` is too generic. Can we use something like `FileChangedChecker` ? It is ugly and long but much more clear.

doc/cmd_run.rst
```diff
@@ -271,6 +271,42 @@ from a task while it is being executed,
 this is controlled by the ``verbosity`` param.
 
 
+check_file_uptodate
+-------------------
+
+`doit` provides different options to check if files are up to date.
+Use the option ``--check_file_uptodate`` to choose:
+
+ * default: use the timestamp, size and md5sum of the file.
```
schettino72: I think we could say that the default is md5sum, the usage of timestamp and size is just an implementation detail for speed optimization. So just say it is md5 and then put a note about the optimization.

doit/cmd_run.py
```diff
@@ -168,31 +183,42 @@ def _execute(self, outfile,
             # user defined class
             reporter_cls = reporter
 
+        # check_file_uptodate
+        if isinstance(check_file_uptodate, six.string_types):
+            if check_file_uptodate not in CHECKERS:
+                msg = ("No check_file_uptodate named '%s'."
+                       " Type 'doit help run' to see a list "
+                       "of available checkers.")
+                raise InvalidCommand(msg % check_file_uptodate)
```
schettino72: I know this is copy and paste from another part of the code but when adding new code please prefer the use of `{}`  and `format`

schettino72:
Excellent patch. thanks.

I would like only to consider the case of users switching back and forth the usage of different checkers.
We need to specify and document this behaviour. Otherwise just some minor inline comments. I still didnt try to run it on my machine so more things might come later...

doit/dependency.py
```diff
@@ -551,18 +608,18 @@ def get_status(self, task, tasks_dict):
 
 class JsonDependency(DependencyBase):
     """Task dependency manager with JSON backend"""
-    def __init__(self, name):
-        DependencyBase.__init__(self, JsonDB(name))
+    def __init__(self, name, **kwargs):
```
schettino72: I would prefer to avoid `kwargs` here. just put the name of the parameter!
Using the name makes it more readable.

schettino72:
@saimn are you planning to also implement this checker configuration option on task level ?

doit/dependency.py
saimn: Done.

doit/dependency.py
saimn: Bad paste, sorry.

doit/dependency.py
saimn: Done.

saimn:
> I would like only to consider the case of users switching back and forth the usage of different checkers.

Yeah, that's indeed a good point and I don't know yet how to handle this :/. If the state format changes, it will break at the some point, depending on the backend ... That's why I kept the same format for the timestamp checker :).
So I think at least I need to find a way to know when the state changed, and if this happens the user has to forget and re-run the tasks ? Then it would be great to have a cmd to tell doit to recompute the dependency states (without running the actions), this command would be useful in other cases too, when you work on your dependency tree for example.

> @saimn are you planning to also implement this checker configuration option on task level ?

I don't think I need it, at least for now, so unless it is easy to do I don't think I will do it. Or maybe later, if there is a need for this.

doit/dependency.py
saimn: I did it this way because in the md5 checker there is a call to get the current state.
So instead I know pass the current state to `get_state`.

saimn:
> at least I need to find a way to know when the state changed, and if this happens the user has to forget and re-run the tasks ?
I have added something for this in save_success / get_status. I save the name of the checker, and return the 'run' state if the checker changes.

doit/dependency.py
```diff
@@ -361,6 +338,94 @@ def remove_all(self):
...
+        @returns (bool): True if dep is modified
+        """
+        pass
+
+    def get_state(self, task, dep, current_state):
+        """Compute the state of a task after it has been successfuly executed.
+
+        @param task (Task): the task object.
+        @param dep (str): path of the dependency file.
+        """
```
schettino72: Please document the return value. Including the special handling for None.

doit/dependency.py
```diff
@@ -361,6 +338,94 @@ def remove_all(self):
...
+        @param state (tuple): state that was previously saved with
+            ``get_state``
+        @returns (bool): True if dep is modified
+        """
+        pass
+
+    def get_state(self, task, dep, current_state):
+        """Compute the state of a task after it has been successfuly executed.
+
+        @param task (Task): the task object.
```
schettino72: Note that `task` param is not being used. Do you think it will ever be useful?

doc/cmd_run.rst
```diff
@@ -271,6 +272,42 @@ from a task while it is being executed,
...
 
 
+check_file_uptodate
+-------------------
+
+`doit` provides different options to check if files are up to date.
+Use the option ``--check_file_uptodate`` to choose:
+
+ * md5: use the md5sum of the file.
+ * timestamp: use only the timestamp and size.
```
schettino72: use only timestamp, no size. I guess most people would expect something like on `make` where target timestamp is compared. That's not the case, so please mention exactly how it works. i.e. any timestamp change counts as a change in the file_dep.

doit/dependency.py
```diff
@@ -527,23 +588,27 @@ def get_status(self, task, tasks_dict):
                 task.dep_changed = list(task.file_dep)
                 return 'run'
 
+        # check for modified file_dep checker
```
schettino72: I would implement in a different way but I think this way is good because it can also handle the case where each task uses a different Checker (in case it gets implemented in the future).

doit/dependency.py
```diff
@@ -409,21 +475,16 @@ def save_success(self, task):
                 self._set(task.name, "result:", get_md5(task.result))
 
         # file-dep
+        self._set(task.name, 'checker:', self.checker.__class__.__name__)
```
schettino72: Do you think it is possible (and desirable) to not save any name for the default `md5`?
I am afraid of some bloat in the `.doit.db`

schettino72:
I still didnt run the code but I very pleased with the code :) Dont forget to add your name on AUTHORS file. When you think it is ready for merging, mention it here so I can do my final review before merging it.

schettino72:
100% unit-test coverage is a must! If you have problems with that I can handle it later but I would appreciate if you add the missing tests.

doit/dependency.py
saimn: Ah indeed, I think we can remove it, the dep file should be enough to compute the state.

doit/dependency.py
saimn: I didn't looked at the db stuff, so it must be possible but if the checker changes from timestamp to the default one, the 'checker:' column will exists so what must be done then ? Is there a way to delete a column in the db backend ? So, is it really a useful optimization to not save this value for the default checker ?

saimn:
I have added some tests to reach 100% coverage, so I think it's ok now. Also thanks for your time and guiding, I understand much better how doit works now :-).

schettino72:
Thanks a lot. this is merged now (I squashed your commits). There were a few issues (I already fixed them):

- I want doit to always give an error if a file_dep doesnt exist. So i moved this out of the FileChangedChecker. Then for performance reasons I had to pass the file_stat as a parameter because checking if a file exist is done implicit by getting the file stat.

- Other doit commands like `auto` also need to know the checker being used, so I moved the option to apply to any command that sub-class DoitCmd not only `run`. Some commands like `forget` and `clean` dont really need to know about the checker but it was easier to implement this way. 
  I also did some long cleanup to separate clearly some test code that was mixed with production code.

- There was another small issue when changing the checkers. Because MD5 checker can use old values when saving the md5 i needed to explicitly remove all saved info when changing checkers.  

Please review and test my changes.

saimn:
Great, thanks !

Issue #25, opened 2015-02-17
Add a command to recompute dependencies state.


saimn:
When changing `check_file_uptodate` the states of dependencies are dropped, so all the tasks are considered out-of-date. This is an issue when the tasks takes a long time to run, and if the file deps  / targets are already existing. 
So I would like to add a command which recompute the states of dependencies without executing the tasks. I think I will take example on the forget command, to find all tasks/subtasks and call `self.dep_manager.save_success` for each task. Seems reasonable ?
Also suggestions for the cmd name are welcome ;-), something like `rebuilddb` ?

schettino72:
I guess it is reasonable. You just need to pay attention that tasks save other values not only information about file_dep... Worse than that if you just call `save_success()` you will remove previously saved values, so I would not use this function.

The command name should not be called `rebuilddb` because it is not being completely rebuilt!
`set-file-dep-as-uptodate` is the best description but too long so what about `reset-dep`?

saimn:
Ah yes indeed, so I need to update the checker class name, the list and the state of file deps. `reset-dep` seems ok, thanks.

PR #28, opened 2015-02-25
Add a command to recompute dependencies state

saimn:
Ref #25

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,7 @@ Changes
 ======
 
  - fix GH-#22: Allow to customize how file_dep are checked
+ - GH-#25: Add a `reset-dep` command to recompute dependencies state
  - BACKWARD INCOMPATIBLE: signature for custom DB backend changed
 
 
--- a/doit/cmd_help.py
+++ b/doit/cmd_help.py
@@ -106,7 +106,10 @@ class Help(DoitCmdBase):
         print('')
         print("Commands")
         for cmd in sorted(six.itervalues(cmds), key=attrgetter('name')):
-            six.print_("  doit %s \t\t %s" % (cmd.name, cmd.doc_purpose))
+            # the command name is padded with spaces and must be 16 character
+            # wide at maximum (this corresponds to the length of
+            # 'help <task-name>')
+            six.print_("  doit {:16s}  {}".format(cmd.name, cmd.doc_purpose))
         print("")
         print("  doit help              show help / reference")
         print("  doit help task         show help on task dictionary fields")
--- a/doit/cmd_resetdep.py
+++ b/doit/cmd_resetdep.py
@@ -1,5 +1,6 @@
 from .cmd_base import DoitCmdBase, check_tasks_exist
 from .cmd_base import subtasks_iter
+from .dependency import DependencyException
 
 
 class ResetDep(DoitCmdBase):
@@ -46,15 +47,19 @@ to a task that has already run.
         for task in task_list:
             # Save this now because dep_manager.get_status will remove the task
             # from the db if the checker changed.
-            task.values = self.dep_manager.get_values(task.name)
-            task.results = self.dep_manager.get_results(task.name)
+            values = self.dep_manager.get_values(task.name)
+            result = self.dep_manager.get_result(task.name)
 
             try:
                 run_status = self.dep_manager.get_status(task, tasks)
-            except Exception:
+            except DependencyException as e:
                 # Skip exception when a depencency file is missing, and force
                 # the state computation
-                run_status = 'run'
+                write("failed {} ({})\n".format(task.name, str(e)))
+                continue
+
+            task.values = values
+            # task.result = result
 
             # An 'up-to-date' status means that it is useless to recompute the
             # state: file deps and targets exists, the state has not changed,
@@ -63,12 +68,9 @@ to a task that has already run.
                 write("skip {}\n".format(task.name))
                 continue
 
-            try:
-                self.dep_manager.save_success(task)
-            except OSError as e:
-                write("failed {} (missing file_dep: {})\n".format(
-                    task.name, str(e)))
-            else:
-                write("processed {}\n".format(task.name))
+            self.dep_manager.save_success(task)
+            # Store the result directly, so that it is not hashed a second time
+            self.dep_manager._set(task.name, "result:", result)
+            write("processed {}\n".format(task.name))
 
         self.dep_manager.close()
--- a/doit/cmd_resetdep.py
+++ b/doit/cmd_resetdep.py
@@ -1,0 +1,42 @@
+from .cmd_base import DoitCmdBase, check_tasks_exist
+from .cmd_base import subtasks_iter
+
+
+class ResetDep(DoitCmdBase):
+    name = "reset-dep"
+    doc_purpose = "recompute the state of file dependencies"
+    doc_usage = "[TASK ...]"
+    doc_description = None
+
+    cmd_options = ()
+
+    def _execute(self, pos_args=None):
+        filter_tasks = pos_args
+
+        # dict of all tasks
+        tasks = dict([(t.name, t) for t in self.task_list])
+
+        if filter_tasks:
+            # list only tasks passed on command line
+            check_tasks_exist(tasks, filter_tasks)
+            # get task by name
+            task_list = []
+            for name in filter_tasks:
+                task = tasks[name]
+                task_list.append(task)
+                task_list.extend(subtasks_iter(tasks, task))
+        else:
+            task_list = self.task_list
+            # self.outstream.write("processing all tasks\n")
+
+        for task in task_list:
+            run_status = self.dep_manager.get_status(task, tasks)
+            if run_status == 'up-to-date':
+                self.outstream.write("skip %s\n" % task.name)
+                continue
+
+            task.values = self.dep_manager.get_values(task.name)
+            self.dep_manager.save_success(task)
+            self.outstream.write("processed %s\n" % task.name)
+
+        self.dep_manager.close()
--- a/doit/cmd_resetdep.py
+++ b/doit/cmd_resetdep.py
@@ -27,16 +27,28 @@ class ResetDep(DoitCmdBase):
                 task_list.extend(subtasks_iter(tasks, task))
         else:
             task_list = self.task_list
-            # self.outstream.write("processing all tasks\n")
 
+        write = self.outstream.write
         for task in task_list:
-            run_status = self.dep_manager.get_status(task, tasks)
+            # Save this now because dep_manager.get_status will remove the task
+            # from the db if the checker changed.
+            task.values = self.dep_manager.get_values(task.name)
+            task.results = self.dep_manager.get_results(task.name)
+
+            try:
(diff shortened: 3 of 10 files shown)


doit/cmd_resetdep.py
```diff
@@ -0,0 +1,50 @@
+from .cmd_base import DoitCmdBase, check_tasks_exist
+from .cmd_base import subtasks_iter
+
+
+class ResetDep(DoitCmdBase):
+    name = "reset-dep"
+    doc_purpose = "recompute the state of file dependencies"
```
schettino72: what do you think of: "recompute and save the state of file dependencies without executing actions"

doit/cmd_resetdep.py
```diff
@@ -0,0 +1,50 @@
...
+    doc_usage = "[TASK ...]"
+    doc_description = None
```
schettino72: Please add a long description mentioning in which situation this command is useful.

doit/cmd_resetdep.py
```diff
@@ -0,0 +1,50 @@
...
+                task_list.append(task)
+                task_list.extend(subtasks_iter(tasks, task))
+        else:
+            task_list = self.task_list
+
+        write = self.outstream.write
+        for task in task_list:
+            try:
+                run_status = self.dep_manager.get_status(task, tasks)
+            except Exception:
```
schettino72: Every `pass` in exception handling deserves a comment :)

doit/cmd_resetdep.py
```diff
@@ -0,0 +1,50 @@
...
+                run_status = 'run'
+
+            if run_status == 'up-to-date':
+                write("skip %s\n" % task.name)
+                continue
+
+            task.values = self.dep_manager.get_values(task.name)
```
schettino72: I guess you also need to handle `task.result`

doit/cmd_resetdep.py
```diff
@@ -0,0 +1,50 @@
...
+            try:
+                self.dep_manager.save_success(task)
+            except OSError:
+                write("skip (missing file_dep) %s\n" % task.name)
```
schettino72: You used the word `skip` for both:
  - unsuccessful operation (fail because there is no file)
  - success nothing to be done

Could we say? "failed - missing file_dep, can not set bla bla bla"

tests/conftest.py
```diff
@@ -1,19 +1,20 @@
-import os
```
schettino72: Looks like your editor "automatically" messed up the imports.

tests/conftest.py
```diff
@@ -95,6 +96,16 @@ def remove_depfile():
 
     return dep_file
 
+
+# test parametrization, execute tests for all DB backends.
+# create a separate fixture to be used only by this module
```
schettino72: oops not "used only by this module" anymore

tests/test_cmd_resetdep.py
```diff
@@ -0,0 +1,76 @@
...
+from doit.cmd_resetdep import ResetDep
+from doit.dependency import TimestampChecker
+from doit.exceptions import InvalidCommand
+from doit.task import Task
+from tests.conftest import tasks_sample, CmdFactory
+
+
+class TestCmdResetDep(object):
+
+    def test_execute(self, pdepfile, dependency1):
```
schettino72: Why do you think it is required to execute these tests with all backends?
I guess just one should be enough...

schettino72:
thanks for the patch. I just did a quick review. I will take a closer look on weekend...

please add :
- an entry on CHANGES file
- docs

tests/test_cmd_resetdep.py
```diff
@@ -0,0 +1,76 @@
...
+        assert sorted(expected) == sorted(got)
+
+    def test_file_dep(self, pdepfile, dependency1):
+        my_task = Task("t2", [""], file_dep=['tests/data/dependency1'])
+        output = StringIO()
+        cmd_list = CmdFactory(ResetDep, outstream=output, task_list=[my_task],
+                              dep_manager=pdepfile)
+        cmd_list._execute()
+        got = output.getvalue()
+        assert "processed t2\n" == got
```
schettino72: Apart from checking the output "saying" the new value was saved, can it also check that the new value was actually saved ?

schettino72:
Please updated the files `bash_completion_doit` and `zsh_completion_doit`. These files are generated using `doit tabcompletion`

saimn:
Hi @schettino72, 
Sorry for the delay before getting back to this PR, I was busy with other things. I have updated the code to take care of all your comments (or at least I hope so :)).

doit/dependency.py
```diff
@@ -502,6 +501,12 @@ def get_value(self, task_id, key_name):
             raise Exception(msg % (task_id, key_name))
         return values[key_name]
 
+    def get_results(self, task_name):
+        """get all saved results from a task
```
schettino72: Each task has only 1 `result`, so it doesnt make sense to say "get all". And the function should be called `get_result()`.

doit/cmd_resetdep.py
```diff
@@ -0,0 +1,74 @@
...
+            for name in filter_tasks:
+                task = tasks[name]
+                task_list.append(task)
+                task_list.extend(subtasks_iter(tasks, task))
+        else:
+            task_list = self.task_list
+
+        write = self.outstream.write
+        for task in task_list:
+            # Save this now because dep_manager.get_status will remove the task
```
schettino72: good point. can you add a note on get_status() docstring mentioning this side-effect.

doit/cmd_resetdep.py
```diff
@@ -0,0 +1,74 @@
...
+            # from the db if the checker changed.
+            task.values = self.dep_manager.get_values(task.name)
```
schettino72: I guess it would be clear if you save this in a temp variable and only set on `task` object if you really gonna save it.

CHANGES
```diff
@@ -7,6 +7,7 @@ Changes
 ======
 
  - fix GH-#22: Allow to customize how file_dep are checked
+ - GH-#25: Add a command to recompute dependencies state
```
schettino72: Please mention the command name

doit/cmd_resetdep.py
```diff
@@ -0,0 +1,74 @@
...
+            task.results = self.dep_manager.get_results(task.name)
+
+            try:
+                run_status = self.dep_manager.get_status(task, tasks)
+            except Exception:
```
schettino72: Please catch a more specific exception.

doit/cmd_resetdep.py
```diff
@@ -0,0 +1,74 @@
...
+                task_list.extend(subtasks_iter(tasks, task))
+        else:
+            task_list = self.task_list
+
+        write = self.outstream.write
+        for task in task_list:
+            # Save this now because dep_manager.get_status will remove the task
+            # from the db if the checker changed.
+            task.values = self.dep_manager.get_values(task.name)
+            task.results = self.dep_manager.get_results(task.name)
```
schettino72: It is `task.result` not `task.results`!

tests/test_cmd_resetdep.py
```diff
@@ -0,0 +1,98 @@
...
+        my_task.result = "result"
+        my_task.values = {'x': 5, 'y': 10}
+        depfile.save_success(my_task)
+        depfile.checker = TimestampChecker()  # trigger task update
+        output = StringIO()
+        cmd_list = CmdFactory(ResetDep, outstream=output, task_list=[my_task],
+                              dep_manager=depfile)
+        cmd_list._execute()
+        got = output.getvalue()
+        # ipdb> from doit.dependency import DbmDependency
```
schettino72: remove debug statements

tests/test_cmd_resetdep.py
```diff
@@ -0,0 +1,98 @@
...
+from doit.task import Task
+from tests.conftest import tasks_sample, CmdFactory, get_abspath
+
+
+class TestCmdResetDep(object):
+
+    def test_execute(self, depfile, dependency1):
+        output = StringIO()
+        tasks = tasks_sample()
+        cmd_list = CmdFactory(ResetDep, outstream=output, task_list=tasks,
```
schettino72: I guess you used test_cmd_list as a base for this file :)
Please replace `cmd_list` with `cmd_reset`...

tests/test_cmd_resetdep.py
```diff
@@ -0,0 +1,98 @@
...
+        output = StringIO()
+        cmd_list = CmdFactory(ResetDep, outstream=output, task_list=[my_task],
+                              dep_manager=depfile)
+        cmd_list._execute()
+        got = output.getvalue()
+        # ipdb> from doit.dependency import DbmDependency
+        # ipdb> db = DbmDependency(depfile.name)
+        assert "processed t2\n" == got
+        assert {'x': 5, 'y': 10} == depfile.get_values(my_task.name)
+        assert get_md5('result') == depfile.get_results(my_task.name)
```
schettino72: if this test was for real it should fail with current implementation. You need to recreate a different instance of `t2` to test the reset-dep command.

doit/cmd_resetdep.py
schettino72: Actaully I dont get this. If it fails here it will also fail when you call `save_success()` so you could write the error message and `continue` from here.

schettino72:
Thanks for your work. No worries about taking a long time :)

Please also update(merge) your branch with master (but dont squash your commits)

tests/test_cmd_resetdep.py
saimn: Indeed, it was working thanks to the typo on `task.results` :/
Now it fails because `self.dep_manager.get_results` returns the hash, so `save_success` saves the hash of the hash.

doit/cmd_resetdep.py
saimn: Hmm, it seems you're right, I don't remember why I ended with this but I can simplify this.

doit/cmd_help.py
```diff
@@ -106,7 +106,10 @@ def print_usage(cmds):
         print('')
         print("Commands")
         for cmd in sorted(six.itervalues(cmds), key=attrgetter('name')):
-            six.print_("  doit %s \t\t %s" % (cmd.name, cmd.doc_purpose))
+            # the command name is padded with spaces and must be 16 character
```
schettino72: it wont look that good but it is ok if the command has more that 16 chars.

schettino72:
merged. github doesnt detected the merge because I squashed your commits...

thanks for the high quality patches!
I did some minor changes. please check.

saimn:
Looks good, thanks for the thorough review ! :)

schettino72:
@saimn sometimes I feel bad for being so picky on reviews hehehe

By the way, if you wish, you are welcome to get more involved with the project reviewing my commits and others. And I am looking forward for more contributions :)

saimn:
> @saimn sometimes I feel bad for being so picky on reviews hehehe

It is not a bad thing :D, I tend to do the same on my projects, and doit code base is quite complex with a lot of different use cases, so it's hard for a new contributor to understand every piece of code.
I still have several things I would like to contribute, so I will try to get more involved if time permits.

PR #31, opened 2015-03-21
Utility for %doit IPython magic function.

ankostis:
Inspired from the promoted (software-carpentry's lesson)[http://www.software-carpentry.org/v5/intermediate/doit/01-doit_basics.html] i crafted a really small but usable utility function that registers a ``%doit`` magic function in IPython interactive shells.

PR #34, opened 2015-03-24
GH-#31: Store db-files into ipython profile-dirs.

ankostis:
Override db-files location inside ipython-profile dir, which is certainly writable.

### DIFF
--- a/doit/tools.py
+++ b/doit/tools.py
@@ -290,4 +290,26 @@ def register_doit_as_IPython_magic():  # pragma: no cover
         """
         ip = get_ipython()
         commander = DoitMain(ModuleTaskLoader(ip.user_module))
-        commander.run(line.split())
+
+        ## Override db-files location inside ipython-profile dir,
+        #    which is certainly writable.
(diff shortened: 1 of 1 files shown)


ankostis:
It's strange...TravisCI build has pulled and failed for a different commit(@0fa11a8a7851), not the one i pushed @3377a13cf06

schettino72:
yeah. I broke master myself :-1: , already pushed a fix...

I wont merge this PR now. I will work on something to allow the customization of commands in a sane way. This way is not nice (but not your fault, there is no other way as of today).

ankostis:
It might be good idea to also retrofit the Help methods to return those strings instead of printing them directly to to stdout.

schettino72:
So after https://github.com/pydoit/doit/commit/cafd9b5d091aad34be175f57c3a718724c5da257 I could implement this in a sane way: https://github.com/pydoit/doit/commit/0ebdfa39cc87d6433a326b7251c97bf5324283fa

thanks

PR #35, opened 2015-03-24
make config files optional

schwager-hsph:


schettino72:
[ConfigParser.read()](https://docs.python.org/3.4/library/configparser.html#configparser.ConfigParser.read) already ignores non-existent files. The root of the problem was that I was not checking if the section exist before getting it items. Python docs does not mention it raises an exception... so I overlooked it but should be fixed now with https://github.com/pydoit/doit/commit/19503366a5f9fc6686f74cca08a46de4be3cc980

PR #36, opened 2015-03-24
feature: the PythonAction should not wrap any returned TaskFailed or TaskError exceptions

schwager-hsph:
Sometimes, I have actions that have run time-evaluated conditions.
These actions follow the same strategy: try one thing, if it fails, try something else. 
These are implemented as python actions; an example is here: https://bitbucket.org/biobakery/anadama_workflows/src/ec0cb09417f7ba5ec801b9fa021db7d934c7c015/anadama_workflows/usearch.py?at=master#cl-139

If the python action fails, it's usually because a nested CmdAction returned a TaskError or TaskFailed exception. In that case, the action raises a TaskError with the actual error string masked. The raised TaskError just describes the type of object that was returned; it doesn't tell the user how the action failed.

This feature shows the user how it failed if the PythonAction returns a meaningful exception.

### DIFF
--- a/doit/action.py
+++ b/doit/action.py
@@ -390,6 +390,9 @@ class PythonAction(BaseAction):
         elif isinstance(returned_value, dict):
             self.values = returned_value
             self.result = returned_value
+        elif isinstance(returned_value, TaskFailed) \
+             or isinstance(returned_value, TaskError):
+            return returned_value
         else:
             return TaskError("Python Task error: '%s'. It must return:\n"
                              "False for failed task.\n"
--- a/doit/action.py
+++ b/doit/action.py
@@ -390,8 +390,7 @@ class PythonAction(BaseAction):
         elif isinstance(returned_value, dict):
             self.values = returned_value
             self.result = returned_value
-        elif isinstance(returned_value, TaskFailed) \
-             or isinstance(returned_value, TaskError):
+        elif isinstance(returned_value, (TaskFailed, TaskError)):
             return returned_value
         else:
             return TaskError("Python Task error: '%s'. It must return:\n"
--- a/tests/test_action.py
+++ b/tests/test_action.py
@@ -353,6 +353,21 @@ class TestPythonAction(object):
         got = my_action.execute()
         assert isinstance(got, TaskError)
 
+    def test_error_taskfail(self):
+        # should get the same exception as was returned from the
+        # user's function
+        def error_sample(): return TaskFailed("too bad")
+        ye_olde_action = action.PythonAction(error_sample)
+        ret = ye_olde_action.execute()
+        assert isinstance(ret, TaskFailed)
+        assert str(ret).endswith("too bad\n")
+
+    def test_error_taskerror(self):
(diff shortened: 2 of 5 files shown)


schettino72:
Well, `CmdAction` is supposed to be used only by doit itself, not for you to use inside a PythonAction.

You should be using plain stdlib's  subprocess. Any specific reason you chose to use CmdAction over subprocess?

schwager-hsph:
I chose to use `CmdAction` because it does exactly what I need:
* `CmdAction` inspects command exit status and returns exceptions understood by doit upon nonzero exit status
* `CmdAction` substitutes task-level items intelligently into the command to be run

These are features I'd re-implement if I were not able to use doit's `CmdAction`

schwager-hsph:
That is to say, if you have suggestions on how to make conditional actions, I'm all ears.

schettino72:
> CmdAction inspects command exit status and returns exceptions understood by doit upon nonzero exit status

That is trivial and we can expose somehow.

> CmdAction substitutes task-level items intelligently into the command to be run

But you don't even pass the a `task` to the CmdAction, so it wont do anything special.

I am not convinced, hehe. Can you provide me a minimal example ?

schettino72:
Ok. After thinking more about it... Although explicit creation of a CmdAction from a PythonAction wont ever be recommended. If doit ever support some way to to execute commands from a python-action that preserve the verbose settings it will require your patch anyway. So this patch is ok.

Please add:
 * your name on AUTHORS file
 * an entry on CHANGES
 * update docs mentioning that a PythonAction can directly return TaskError/TaskFailed

doit/action.py
```diff
@@ -390,6 +390,9 @@ def execute(self, out=None, err=None):
         elif isinstance(returned_value, dict):
             self.values = returned_value
             self.result = returned_value
+        elif isinstance(returned_value, TaskFailed) \
```
schettino72: The second argument of `isinstance()` accepts a tuple of types. https://docs.python.org/3/library/functions.html#isinstance

tests/test_action.py
```diff
@@ -353,6 +353,21 @@ def error_sample(): return object()
         got = my_action.execute()
         assert isinstance(got, TaskError)
 
+    def test_error_taskfail(self):
+        # should get the same exception as was returned from the
+        # user's function
+        def error_sample(): return TaskFailed("too bad")
+        ye_olde_action = action.PythonAction(error_sample)
```
schettino72: what is the meaning of "ye olde" ?

schettino72: ok, got it. :+1:  http://en.wikipedia.org/wiki/Ye_Olde

I was thinking it was some kind of acronym... you can leave it this way :)

schwager-hsph:
Ok! I've made the proposed changes to the branch. Please let me know if there's more I can / should do!

schettino72:
Merged. thanks! github doesnt detect because i squashed your commits.

Please pay attention not to add white-space at the end of lines... I had to clean up several of those

PR #37, opened 2015-03-24
Feature: CmdParse can construct lists by providing an 'append' style option

schwager-hsph:
Akin to the optparse 'append' action (https://docs.python.org/2/library/optparse.html#other-actions).

Providing multiple flags on the command line constructs a list.
For example, `doit -l foo -l bar` translates to `['foo','bar']` for the `-l` flag.

### DIFF
--- a/doit/cmdparse.py
+++ b/doit/cmdparse.py
@@ -4,7 +4,7 @@ Built on top of getopt. optparse can't handle sub-commands.
 """
 import getopt
 import six
-
+import copy
 
 class DefaultUpdate(dict):
     """A dictionary that has an "update_defaults" method where
@@ -209,10 +209,7 @@ class CmdParse(object):
         params = DefaultUpdate()
         # add default values
         for opt in self.options:
-            if opt.type is list:
-                params.set_default(opt.name, list())
-            else:
-                params.set_default(opt.name, opt.default)
+            params.set_default(opt.name, copy.copy(opt.default))
 
         # parse options using getopt
         try:
--- a/doit/cmdparse.py
+++ b/doit/cmdparse.py
@@ -80,7 +80,7 @@ class CmdOption(object):
                 raise CmdParseError(msg % (opt_dict, field))
 
         self.name = opt_dict.pop('name')
-        self.default = opt_dict.pop('default')
+        self._default = opt_dict.pop('default')
         self.type = opt_dict.pop('type', str)
         self.short = opt_dict.pop('short', '')
         self.long = opt_dict.pop('long', '')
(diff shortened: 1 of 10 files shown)


schwager-hsph:
Sorry, didn't mean to add in the PythonAction commit.

tests/test_cmdparse.py
```diff
@@ -153,26 +160,33 @@ def test_getOption(self, cmd):
...
+        assert ['foo', 'bar'] == params['list']
 
     def test_parseLongValues(self, cmd):
-        params, args = cmd.parse(['--rare-bool','--num','89', '--no-flag'])
+        params, args = cmd.parse(['--rare-bool','--num','89', '--no-flag',
+                                  '--list', 'flip', '--list', 'flop'])
         assert True == params['rare']
         assert False == params['flag']
         assert 89 == params['num']
+        assert ['foo', 'bar', 'flip', 'flop'] == params['list']
```
schettino72: looks like a bug that `foo` and `bar`  show up here.

tests/test_cmdparse.py
```diff
@@ -153,26 +160,33 @@ def test_getOption(self, cmd):
...
 
     def test_parsePositionalArgs(self, cmd):
-        params, args = cmd.parse(['-f','p1','p2','--sub-arg'])
```
schettino72: why you removed this?

doit/cmdparse.py
```diff
@@ -213,8 +213,8 @@ def parse(self, in_args):
 
         # parse options using getopt
         try:
-            opts, args = getopt.getopt(in_args, self.get_short(),
-                                       self.get_long())
+            opts, args = getopt.gnu_getopt(in_args, self.get_short(),
```
schettino72: I guess this change will break other parts of the code.

tests/test_cmdparse.py
schwager-hsph: I changed this in a627702 to give `cmd.parse` a list of recognized options. `gnu_getopt` picks up on these later-placed options now.

tests/test_cmdparse.py
schwager-hsph: Fixed in cfd8b62 and 509cb4e. Thanks for the good catch!

schwager-hsph:
Upon further review, I found that I didn't really need `gnu_getopt` in my code, either. I just needed to subclass `doit.cmd_run:Run` correctly. Sorry about that!

schettino72:
please also add an entry on CHANGES, and docs ( I guess the arguments section should mention lists are also acceptable).

doit/cmdparse.py
```diff
@@ -4,7 +4,7 @@
 """
 import getopt
 import six
-
```
schettino72: I tend to use more blank lines than usual. IMO this increases readability... please dont remove my blank lines :D

schwager-hsph: My mistake. I'll be more careful about blank lines :)

schettino72:
merged. thanks!

PR #40, opened 2015-03-26
Py3 tutorial

ankostis:
Make tutorial-code py3 compatible.

- Fix print-statement.
- Drop {}.iterkeys().
- STILL left one `iteritems()` (doc/tutorial/cproject.py)

### DIFF
--- a/doc/tutorial/clean_mix.py
+++ b/doc/tutorial/clean_mix.py
@@ -1,7 +1,8 @@
+from __future__ import print_function
 from doit.task import clean_targets
 
 def simple():
-    print "ok"
+    print("ok")
 
 def task_poo():
     return {
--- a/doc/tutorial/calc_dep.py
+++ b/doc/tutorial/calc_dep.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 DOIT_CONFIG = {'verbosity': 2}
 
 MOD_IMPORTS = {'a': ['b','c'],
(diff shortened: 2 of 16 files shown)


doc/tutorial/clean_mix.py
```diff
@@ -1,7 +1,7 @@
 from doit.task import clean_targets
 
 def simple():
-    print "ok"
```
schettino72: it seems you removed all quotes `"`

ankostis: Crafting untested commits from regexes late at night will not work as intended :-(
Wil fix them manually.

saimn: python-modernize is quite handy for this, for example for prints you can use this:
```
python-modernize -w -f libmodernize.fixes.fix_print dodo.py
```
http://python-modernize.readthedocs.org/en/latest/

ankostis: great thanks!

ankostis:
Take 2: Used `future` modernize with hand-removing needless changes.

Left ``from __future import print`` statement in case the tutorial-code gets test-cased ever.

schettino72:
thanks!

Kwpolska:
In most cases, you could leave the __future__ import out, provided that you don’t use commas.

```pycon
>>> #py2
>>> print("hello")
hello
>>> print("hello", "world")
("hello", "world")
```

```pycon
>>> #py3
>>> print("hello")
hello
>>> print("hello", "world")
hello world
```

PR #41, opened 2015-03-26
FIX #38: Minor err-msg about magic-args mismatch.

### LINKED ISSUE #38: cmd-action substitution keys should include (almost) all task properties
The most important task-properties are the "dependency fields" and these are good to have them in the substitution-dictionary for debugging purposes.
Currently even a simple `echo %(name)s` cmd-action does not work.

As a side note, it would facilitate the user if at action.py@, ine:248 it catches `KeyError` and prints the original cmd's format-text causing the problem.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10015944-cmd-action-substitution-keys-should-include-almost-all-task-properties?utm_campaign=plugin&utm_content=tracker%2F1885485&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F1885485&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ankostis:
Note that py2 tests/test_filewatch.py fails on my sytem (and it did before my PR).

### DIFF
--- a/doit/action.py
+++ b/doit/action.py
@@ -180,6 +180,8 @@ class CmdAction(BaseAction):
         """
         try:
             action = self.expand_action()
+        except InvalidTask:
+            raise
         except Exception as exc:
             return TaskError("CmdAction Error creating command string", exc)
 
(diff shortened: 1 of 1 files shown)


schettino72:
Great, but it need tests. 

Also prefer to use `"".format()` instead of interpolation with `%` for new code.

schettino72:
> Note that py2 tests/test_filewatch.py fails on my sytem (and it did before my PR).

This test is flaky/unstable. On my machine it always work but on travis it is also unstable.

This test is horrible, and complicated. And I wonder if I should just disable it by default...

schettino72:
Closing because inactive for a long time. Please re-open if you work on it again.

PR #42, opened 2015-03-31
Load config ini version compatibility issue

schwager-hsph:
Python 2's ConfigParser doesn't have the 'delimiters' keyword. This patch navigates around that problem.


schettino72:
Check `setup.py` (or even README) a backport of python3's [configparser] (https://pypi.python.org/pypi/configparser) is required for python2.

PR #43, opened 2015-03-31
Implement Cmdparse choices options

schwager-hsph:


schettino72:
Thanks for working on this.

First: please `rebase` your branch to master. Just `git rebase master` should do the magic.

I dont get why choices is a `dict` it should be a tuple (or a set), the meaning of the string is up to the command, cmdparse should not know or care at all about it.

schwager-hsph:
So that's what `rebase` is for. Thanks! 

Choosing `dict` as the type for choices options was merely the first idea that occurred to me.
I don't understand what you mean when you say "cmdparse should not know or care at all about [the meaning of the string]." 
I think using a dictionary simplifies what everyone would use a choices-style option for. It generalizes away any choice-switching code. i.e. the user won't have to write this:
```
def my_action(some_choice_str):
    if some_choice_str == "a":
        do_a()
    elif some_choice_str == "b":
        do_b()
#etc
```
And the user won't even need to write this:
```
def my_action(some_choice_str):
    some_func = CHOICE_MAP[some_choice_str]
    some_func()
```

schwager-hsph:
That said, you're the owner of this repository. I'll refactor if you want to do this with a `set`.

schettino72:
Please make changes only on cmdparse for now. Do not bother to convert all options that doit uses that could be a CHOICE to use the feature just now. You were breaking some code...  and I plan to some work related to that with plugins soon.

doit/cmdparse.py
```diff
@@ -147,7 +156,16 @@ def help_doc(self):
 
         text = []
         opt_str = self.help_param()
-        opt_help = self.help % {'default': self.default}
+
+        if isinstance(self.type, dict):
+            default = next(iter( key for key in six.iterkeys(self.type)
```
schettino72: thats one problem of using a dict

doit/cmdparse.py
```diff
@@ -147,7 +156,16 @@ def help_doc(self):
...
+                                 if self.type[key] == self.default ))
+            template_dict = {'choices': self.type.keys(),
```
schettino72: lets use `format()` instead of string interpolation `%`.

Since the only value that could appear before was `%(default)s` we can do:

```
if "%(default)s" in self.help:
    # print deprecation warning
    opt_help = self.help % {'default': self.default}
else:
    opt_help = self.help.format(choices=choices, default= self.default) 
```

doit/cmd_completion.py
```diff
@@ -7,13 +7,19 @@
...
+}
+
 opt_shell = {
     'name': 'shell',
     'short': 's',
     'long': 'shell',
-    'type': str,
-    'default': 'bash',
-    'help': 'Completion code for SHELL. default: "bash". options: [bash, zsh]',
+    'type': SHELL_CHOICES,
```
schettino72: this one you can use the CHOICES implementation if you wish.

schettino72:
> I don't understand what you mean when you say "cmdparse should not know or care at all about [the meaning of the string]." Perhaps you can point me to specific lines of code?

Using a dict you are assuming that the choices are a kind of mapping but thats not always the case.
But it is more like my design style. Other argument parsers let you even attach callables or define the an operation to be done for a given argument. doit parser doesnt allow that, it is minimal in a sense that it only tells if an argument was present or not and its value.

schwager-hsph:
I remember now another reason for choosing a dictionary instead of a set or tuple. Using a dictionary for `opt_parallel_type` in cmd_run.py line 101 lets me use my own `Runner` in AnADAMA without copying code or monkey patching. I can just import the dictionary and insert my custom runner: https://bitbucket.org/biobakery/anadama/src/847b2cb8a6c31afa62bebad92d30d045f753177e/anadama/runner.py?at=doitchanges#cl-77

The reasoning above is really why I'm being a stick in the mud about using a dictionary. Plus, I'd like to argue that a dictionary isn't incompatible with your style:
* A dictionary can map to anything, including strings as seen in the cmd_completion.py commit above
* A dictionary simplifies other parts of the code in many places c.f. 5c627932 and 8eaf61bb6
* At worst, a dictionary mapping makes the code only slightly more complex c.f. https://github.com/pydoit/doit/pull/43/files#diff-41073979ec71645a98837911730ae4b9L49

schettino72:
Your "simplification" breaks code. i.e. https://github.com/pydoit/doit/blob/master/doc/tutorial/custom_reporter.py
Note how the value for a reporter in DOIT_CONFIG can be a string or a reporter class.

We are adding plugins, so dont worry about your custom runner. We gonna tackle that soon...

schettino72:
One more thing. Each `choice` should be a two-value tuple with *name* and *description*, *description* is not really required but often used when displaying help.

schwager-hsph:
I'll redo the pull request so it's less messy.

PR #44, opened 2015-03-31
Fix the list -s command when result_dep is used.

saimn:
The tasks dict was not passed to `get_status` when using `doit list --all -s mytask`

```
  File "/home/conseil/lib/doit/doit/doit_cmd.py", line 175, in run
  File "/home/conseil/lib/doit/doit/cmd_base.py", line 111, in parse_execute
  File "/home/conseil/lib/doit/doit/cmd_base.py", line 352, in execute
  File "/home/conseil/lib/doit/doit/cmd_list.py", line 149, in _execute
    self._print_task(template, task, status, list_deps)
  File "/home/conseil/lib/doit/doit/cmd_list.py", line 80, in _print_task
    task_status = self.dep_manager.get_status(task, None)
  File "/home/conseil/lib/doit/doit/dependency.py", line 576, in get_status
    uptodate_result = utd(*args, **utd_kwargs)
  File "/home/conseil/lib/doit/doit/task.py", line 537, in __call__
    dep_task = self.tasks_dict[self.dep_name]
TypeError: 'NoneType' object has no attribute '__getitem__'
```

### DIFF
--- a/doit/cmd_list.py
+++ b/doit/cmd_list.py
@@ -68,7 +68,7 @@ class List(DoitCmdBase):
     STATUS_MAP = {'ignore': 'I', 'up-to-date': 'U', 'run': 'R'}
 
 
-    def _print_task(self, template, task, status, list_deps):
+    def _print_task(self, template, task, status, list_deps, tasks):
         """print a single task"""
         line_data = {'name': task.name, 'doc':task.doc}
         # FIXME group task status is never up-to-date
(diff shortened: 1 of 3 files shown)


schettino72:
Thanks. This is merged now. 

I did not use your tests because I dont want to let "sample_tasks" grow more since it affects many tests. It is ok to use only on very basic tests...

Issue #20, opened 2015-01-22
specify target of a DelayedTask on command line


schettino72:
Since not all tasks are created before execution starts, it is required some special handling for target names of targets that have not been created yet.

See discussion on https://github.com/getnikola/nikola/issues/1562#issuecomment-70836094

General idea is that if a target is not found, before raising an error to the user try to load DelayedTasks (as done by command list) to look for the given `target` name.

Some considerations:

 1) as of now a DelayedTask creates an implicit `task_dep` for the task given in it's `executed` param. But this `task_dep` is not preserved when the DelayedTask is re-created. It should not only be preserved but all created tasks should include this `task_dep` because the guarantee that the dependent task was already executed wont exist anymore!

2) if the selected tasks for execution includes know tasks and targets they should be executed **before**, any DelayedTask is loaded to look for an unknown `target`. This will ensure that same command line will work nicely even if it is on its first execution.

felixfontein:
Re 1): very good point.

Re 2): I guess one exception is that one of these tasks depends on a DelayedTask. So something like:
  - first determine which of the tasks and targets specified on the command line (or in ```default_tasks```) is not delayed; start executing as if the command line only containes these;
  - if during execution of these and their prerequisites one of the "missing" tasks/targets shows up, also make sure it and its prerequisites is executed;
  - finally, if all tasks are executed but some tasks/targets haven't been found, begin loading DelayedTasks until something is found (if yes, continue with it, etc.);
  - when everything specified on the command line is done, stop;
  - if doit runs out of DelayedTasks before this happens, terminate with error (as before).

Is this more or less what you meant?

schettino72:
yes. thats it. 

Lets use the same dodo.py file as example to test this feature. https://github.com/pydoit/doit/blob/master/doc/tutorial/delayed.py should be good enough.

Issue #30, opened 2015-03-21
Delayed task is never executed if basename is set


productivememberofsociety666:
## Description
If I set the ``"basename"`` option of a task that has been created via ``doit.create_after``, this task is never executed.
## Minimal example to reproduce
Here is a minimal ``dodo.py`` file to reproduce the error. You can see from the output to ``sys.stderr`` that this is not merely a case of not reporting the task's run, but it actually doesn't run at all.
Note that removing either the ``"basename"`` option or the ``create_after`` decorator produces the "correct" behaviour; it is only when used together that they cause trouble.
#### dodo.py:
````python
from doit import create_after
import sys

def say_hello(your_name):
  sys.stderr.write("Hello from {}!\n".format(your_name))

def task_a():
  return {
    "actions": [ (say_hello, [ "a" ] ) ]
  }

@create_after("a")
def task_b():
  return {
    "actions": [ (say_hello, [ "b" ] ) ],
    "basename": "B saying hello"
  }
````
#### Terminal log:
````
$ doit
.  a
Hello from a!
.  B saying hello
Hello from b!
$ 
````
#### Terminal log:
````
$ doit
.  a
Hello from a!
$ 
````

## System information
* *doit* version: 0.27.0
* *Python* version: 3.4.2

**Comment by schettino72** [comment]
The best bug report I ever got :+1: 

I forgot to write a note on docs about this... Since the metadata of a delayed task is only created on demand it is currently not possible to set a `basename`. doit needs to know which task-creator function needs to be execute the get the metadata of the requested task.

What I thought before was to extend the decorator to accept a `creates` argument, something like

```
@create_after("a", creates=['B saying hello'])
def task_b():
   ...
```

schettino72:
I added a note in docs about this limitation: https://github.com/pydoit/doit/commit/7fd9f9ac58a4053dd6b008da984601aa8a540913

productivememberofsociety666:
That was fast - thank you!

PR #46, opened 2015-04-05
Allows to specify target of a DelayedTask on command line.

felixfontein:
First shot at implementing #20: allowing to specify targets and delayed tasks of ```DelayedTask```'s on the command line.

This is based on doit 0.27.0, to be able to test it with Nikola's earlytask_impl branch (see https://github.com/getnikola/nikola/pull/1632).

### DIFF
--- a/doit/control.py
+++ b/doit/control.py
@@ -52,9 +52,18 @@ class TaskControl(object):
             self._def_order.append(task.name)
 
         # expand wild-card task-dependencies
+        unpack_all = False
         for task in six.itervalues(self.tasks):
-            for pattern in task.wild_dep:
-                task.task_dep.extend(self._get_wild_tasks(pattern))
+            if len(task.wild_dep) > 0:
+                unpack_all = True
+                break
+        if unpack_all:
+            # execute all delayed loaders to make sure all tasks are actually there
+            while self._unpack_delayed_task():
+                pass
+            for task in six.itervalues(self.tasks):
+                for pattern in task.wild_dep:
+                    task.task_dep.extend(self._get_wild_tasks(pattern))
 
         self._check_dep_names()
         self.set_implicit_deps(self.targets, task_list)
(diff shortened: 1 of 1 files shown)


felixfontein:
(Sorry it took that long... There were too many other things happening :) )

schettino72:
Hi, thanks for working on this.

I would like to know more about your concrete use-case. For which targets in Nikola you would use this. I am asking this because I am thinking that attaching regexp to identify which targets a task creates might be a better solution.

regexp advantages:
  - result is always correct, even on first run because a task_dep will be executed
  - no need to load DelayedTasks that wont be used
  - cleaner implementation

regexp disavantages:
  - It might not be possible unambiguosly describe task targets by a regexp

I guess that in real life a regexp should be suitable, so lets check your use-case with Nikola :)

Also related to this is #30 (but about basename instead of target)

felixfontein:
Well, there are two main reasons I'd like this feature:

 - More meta: to make nikola's earlytask_impl branch more compatible with current behavior;

 - More conrete: I want to be able to tell nikola to build specific targets, which isn't possible in the earlytask_impl branch as all tasks (which specify the targets) are created by delayed task loaders.

I also don't think this collides with regexps, as there's always the choice to choose ".*" as a regexp describing the generated tasks' targets. (And for some tasks, it might not even be possible to specify something more concrete.)

I think regexp support would be nice to have (as task generators which can specify quite clearly which targets they generate don't have to be loaded), but especially for backwards compatbility with older plugins (or in some cases, when it is not possible to say anything more concrete about targets except that they are in ```output/```, for example) this won't help much.

schettino72:
I agree this feature is useful, what I am saying is that regexp is superior in many aspects and its limitations are not very relevant in practice. I forgot to mention another big regexp disavantage: you need to werite the regexp!

For my own Nikola based blog, I guess it would not be hard to make a regexp for each target.
 - indexes: "/output/index(.*).html"
 - posts: "/output/posts/(.*).html"
 - categories: "/output/categories/(.*).html"
 - archives: "/output/(\d\d\d\d)/(.*).html"

Yes it would affect backward compatibility with plugins but this is also not coming for free...
What makes me unconfortable with this change is that doit might do the right thing *or not* depending on the state of previous runs. It is very bad for  build-tool when people stop trusting it.

schettino72:
I just implemented https://github.com/pydoit/doit/commit/03f0cf7e1b4b6915c35619ceab367bdda77e8efd to allow specifying a **sub-task** from a DelayedTask. It is almost the same as what it would take to implement specifying **targets** (but a bit easier). 

The patch itself made it even more clear to me that use of regexp is the right way to go.
I am sorry if you feel that I mislead you to implement a patch that I wont accept, sometime it is hard to 
 make a call before actually looking at an implementation. Sorry my mistake.

felixfontein:
I still don't see what you want to do in case you find two regexs matching a target, or if a delayed task creates a task which name it didn't announce before. Or is that not allowed? (Which would be ways to strict, from my point of view.)

schettino72:
@felixfontein please check if this branch is good enough for you: https://github.com/pydoit/doit/compare/delayed-target-regex?expand=1

Announcing a name is optional. There is no support for 2 regex matching a target (but could be added).

felixfontein:
I don't think it is always possible to avoid having not two regexps matching the same target. At least, not with an extreme amount of PITA and (even worse) user interaction. In particular for Nikola, I don't want to force end-users to start writing regexps, and I neither want to reduce Nikola's flexibility of allowing to put things where you want them. So I'm still convinced that the code I wrote (or something else doing something similiar) is needed.

schettino72:
The code could be improved to handle more than one matching regexps.

As I said before the problem with your code is that it doesnt work if the targets intermediates were never built. IMO that would make doit not *trustable* ... Any other idea ?

felixfontein:
But how is that different from when you specify a delayed task on the command line, where the targets also have dependencies which are produced by some other delayed task? What does doit currently do in that situation?

PR #47, opened 2015-04-07
Cmdparse choices options tuple

schwager-hsph:
Let's try this feature again.


schettino72:
merged. thanks.

I did a few changes: 
 - choices are always displayed on help, no need to include a `%(choices)s` in your help string
 - added a new `choices` attribute for option dictionary, `type` must be a callable to convert a string to the "real" value
 - more strict on choices specification. must be 2-tuple.

please check and let me know if you have any concerns.

Issue #49, opened 2015-04-22
`auto` does not work on Mac OS X


philipbl:
Adding the `auto` command line option does not work on Mac OS X (10.10.3). The task is run once, but then nothing occurs after that. I have played around with the source code and from what I can tell, the [process that `doit auto` creates](https://github.com/pydoit/doit/blob/master/doit/cmd_auto.py#L105-L107) is not exited when the file system change event is [received](https://github.com/pydoit/doit/blob/master/doit/cmd_auto.py#L93-L95).

Here is a simple `dodo.py` file that shows this problem:

```python
def task_test(): 
    return { 
        'actions': ['echo "test" > test.txt'], 
        'file_dep': ['watch.txt']             
        }
```

If you run `doit auto` and modify `watch.txt` on Mac OS X, nothing happens.

WIP at #101

schettino72:
Thanks for the report. The problem is not so simple.

Check `filewatch.py`, it contains the notifier loop for both linux and mac.  From what I remember it works in a different way than the linux one. 
The linux version would start a new process after every notifcation but the Mac version doesnt.
I guess the difference came from the different API provided by the 2 systems. 

One advantage of creating a new process is that it doesnt cache the python imports.

I did not implement the Mac version myself... and I dont use Mac at all. So it would be great if you can contribute a patch.

philipbl:
Sounds good. I will start working on it.

schettino72:
Or maybe we should just use https://github.com/gorakhargosh/watchdog

I didnt look at it but I hope that using watchdog we could could support linux/mac/bsd/windows easily.

philipbl:
I have a working version of doit that uses watchdog. I need to clean up the code before I submit a PR. Also, I only have access to Mac OS and Linux for testing. It would be great if we could test the new watchdog functionality on Windows and other OSes.

schettino72:
fixed on #135

PR #51, opened 2015-05-01
Update `auto` command to use Watchdog

philipbl:
The `auto` command now uses [watchdog](https://github.com/gorakhargosh/watchdog) instead of pyfsevents and pyinotify. This provides better platform compatibility (and fixes issue #49), though it only has been tested on Linux and Mac OS. Since I have only tested on these two platforms, I did not update the `supported_platforms` variable in `filewatch.py`.

Watchdog might need to be configured differently. Currently, I have it set to only trigger when files are modified and it does not recursively search subdirectories (lines 71 and 80 of `filewatch.py`). All unit tests pass, but I am not sure what the desired behavior of `doit` is.

The `test_watchfile.py` tests have been changed since no callback can be passed in like previous tests.

### DIFF
--- a/doit/cmd_auto.py
+++ b/doit/cmd_auto.py
@@ -92,7 +92,7 @@ class Auto(DoitCmdBase):
             class DoitAutoRun(FileModifyWatcher):
                 def handle_event(self, event):
                     # print("FS EVENT -> {}".format(event))
-                    sys.exit(result)
+                    return False
             file_watcher = DoitAutoRun(watch_files)
             # kick start watching process
             file_watcher.loop()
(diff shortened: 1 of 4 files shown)


schettino72:
thanks @philipbl I might take a few days before I review this...

schettino72:
Hi @philipbl , sorry for the delay.

I changed the code to use a threading lock instead of sleeping: https://github.com/pydoit/doit/commit/eb99ce71d35292cb7b19ffc331f0610d8f448c19
otherwise changes would not be noticed immediately... this change broke tests (i didnt fix them)

A blocker issues is gorakhargosh/watchdog#297 . I guess pip/setuptools does not allow installation without automatically pulling the dependencies.

So we need to convince them to make a separate release of the core lib... can you check with them if they are willing to do that? Worst case scenario we just fork the project and publish just the lib on pypi ourselves. 

Another option is publish the `auto` command separate of `doit` core. In that case I would not bother about the extra dependencies from watchdog.

schettino72:
Another option: https://pypi.python.org/pypi/pywatchman

schettino72:
work is now on PR #101

Issue #50, opened 2015-04-27
Executing tasks in parallel fails on Windows


joschkazj:
First of all a thank you for this tool, which seems to be very useful for automating scientifc workflows. I am using doit for automating a training-evaluation workflow, which contains many embarrassingly parallel tasks so the ``-n <NUM_JOB>`` options is exactly what I want. Unfortunately there are some pickling isses when doing so on Windows.

My software stack:
```
configparser              3.3.0.post2
doit                      0.28.0
pip                       6.1.1
python                    2.7.9
setuptools                15.1
six                       1.9.0
```

`dodo.py` (instead of the `echo` commands, some python scripts are started in my application, but the parallel execution behavior is the same):
```python
# -*- coding: utf-8 -*-
import os
import os.path as osp

from doit import tools

FEATURES = ["lbp_small", "lbp_medium", "lbp_72angles", "hog_normalised",
            "hog_default", "daisy_default", "hog_single_cell"]
OUT = "out"

paths = {}
paths["OUT_FEATURES"] = osp.join(OUT, "features")
paths["OUT_EVALUATION"] = osp.join(OUT, "evaluation")
paths["OUT_FIGURES"] = osp.join(OUT, "figures")

            
def task_feature_extraction():       
    for feat in FEATURES:
        feat_spec = "feat_{}.json".format(feat)
        feat_file = osp.join(paths["OUT_FEATURES"], "feat_{}.hdf5".format(feat))
        yield {"name": feat_file,
               "actions": ["echo extract %s > %s" % (feat_spec, feat_file)],
               "targets": [feat_file],
               "clean": True,
               # force doit to always mark the task
               # as up-to-date (unless target removed)
               'uptodate': [True]}
```

Command: ``doit -n 4`` causes traceback:
```
Traceback (most recent call last):
  File "C:\Anaconda\envs\surface-classification\lib\site-packages\doit\doit_cmd.py", line 165, in run
    return command.parse_execute(args)
  File "C:\Anaconda\envs\surface-classification\lib\site-packages\doit\cmd_base.py", line 122, in parse_execute
    return self.execute(params, args)
  File "C:\Anaconda\envs\surface-classification\lib\site-packages\doit\cmd_base.py", line 405, in execute
    return self._execute(**exec_params)
  File "C:\Anaconda\envs\surface-classification\lib\site-packages\doit\cmd_run.py", line 239, in _execute
    return runner.run_all(self.control.task_dispatcher())
[...]

PR #53, opened 2015-05-22
Fix pickling on windows so that we can spawn new processes

gstorer:
Fix for #50. 

Note that I elected to skip pickling all of dep_manager. I figure that its likely extra non-picklable things could be added here in the future.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,3 +19,4 @@
  * Simon Conseil - contact <at> saimon <dot> org
  * Kostis Anagnostopoulos - ankostis <at> gmail <dot> com
  * Randall Schwager - schwager <at> hsph <dot> harvard <dot> edu
+ * Gerlad Storer - https://github.com/gstorer
--- a/doit/control.py
+++ b/doit/control.py
@@ -262,6 +262,11 @@ class ExecNode(object):
         # generator from TaskDispatcher._add_task
         self.generator = None
 
+    def __getstate__(self):
+        pickle_dict = self.__dict__.copy()
+        pickle_dict['generator'] = None
+        return pickle_dict
+        
     def reset_task(self, task, generator):
         """reset task & generator after task is created by its own `loader`"""
         task.loader = DelayedLoaded
@@ -314,6 +319,10 @@ class TaskDispatcher(object):
 
         self.generator = self._dispatcher_generator(selected_tasks)
 
+    def __getstate__(self):
+        pickle_dict = self.__dict__.copy()
+        pickle_dict['generator'] = None
+        return pickle_dict
 
     def _gen_node(self, parent, task_name):
         """return ExecNode for task_name if not created yet"""
--- a/doit/reporter.py
+++ b/doit/reporter.py
@@ -24,7 +24,12 @@ class ConsoleReporter(object):
         self.show_out = options.get('show_out', True)
         self.show_err = options.get('show_err', True)
         self.outstream = outstream
-
+        
+    def __getstate__(self):
+        pickle_dict = self.__dict__.copy()
+        pickle_dict['outstream'] = None
+        return pickle_dict    
+        
     def write(self, text):
         self.outstream.write(text)
 
--- a/doit/runner.py
+++ b/doit/runner.py
@@ -47,7 +47,11 @@ class Runner(object):
         self.final_result = SUCCESS # until something fails
         self._stop_running = False
 
-
+    def __getstate__(self):
+        pickle_dict = self.__dict__.copy()
+        pickle_dict['dep_manager'] = None
+        return pickle_dict 
+        
     def _handle_task_error(self, node, catched_excp):
         """handle all task failures/errors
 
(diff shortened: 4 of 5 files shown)


schettino72:
thanks.

- please add unittests, checking that `Runner` object pickle should be good enough.
- add your name on authors file

schettino72:
is this patch really enough? python docs says that  `Process.__init__` `target` can not be a method. But you did not modify that.

gstorer:
It would appear to be enough. 

The docs say everything has to be picklable. I think, from my brief eduction on the subject over the last few of days, that the example given is a simplification. Because ```Runner``` is picklable and the method is specified 'normally' everything is ok.

I'll add some tests in the next day or so.

gstorer:
I've had a closer look at some of the existing unit tests. I'd say you pretty much already have coverage for this if the tests are run on Windows. I'll add an explicit pickle one for quick testing on other platforms.

A bigger problem is the non picklable nature of anonymous code. E.g. all the function defined within methods. A lot of tests are failing on Windows because the tests feed these functions in as arguments. 

Would it be better to try and blank all the user specifiable functions (assuming this won't break things further) or change the unit tests not to use them?

The offending functions in test_runner.py (I'm sure there's heaps elsewhere):
https://github.com/pydoit/doit/blob/master/doit/tests/test_runner.py#L337
https://github.com/pydoit/doit/blob/master/doit/tests/test_runner.py#L477
https://github.com/pydoit/doit/blob/master/doit/tests/test_runner.py#L494
https://github.com/pydoit/doit/blob/master/doit/tests/test_runner.py#L602
https://github.com/pydoit/doit/blob/master/doit/tests/test_runner.py#L684
https://github.com/pydoit/doit/blob/master/doit/tests/test_runner.py#L693

schettino72:
> The docs say everything has to be picklable

Which python version are you using? Actually the docs passage I was referring to is only present in python2 docs... maybe this limitation was fixed in python3.


>  I'd say you pretty much already have coverage for this if the tests are run on Windows. I'll add an explicit pickle one for quick testing on other platforms.

Yes. That's the idea, the functionality is already being tested. But we can exercise the code while testing in all platforms.


> A lot of tests are failing on Windows because the tests feed these functions in as arguments.

If we are serious about having Windows as development platform a CI is required.

schettino72:
> Would it be better to try and blank all the user specifiable functions (assuming this won't break things further) or change the unit tests not to use them?

The code already balnks as much as possible (if not let me know the specific case). So we need to fix the tests.

gstorer:
We're using 2.7 here and it seems to work ok. Perhaps the limitation was removed earlier and the docs didn't get updated.

I fixed the links to the breaking functions that I've found in the tests.

For full Windows support it does indeed seem as though some CI is needed. I'm still hoping someone else will volunteer there though...

joschkazj:
Thanks for looking into this.
I've tried the changes locally on Windows and can confirm the issue with the nested functions:

> A bigger problem is the non picklable nature of anonymous code. E.g. all the function defined within methods. A lot of tests are failing on Windows because the tests feed these functions in as arguments.

Additionally the test [TestMRunner_parallel_run_tasks.test_task_not_picklabe_multiprocess](https://github.com/pydoit/doit/blob/master/tests/test_runner.py#L683) (BTW: shouldn't it be called ...picklab**l**e?) fails for me. This can be fixed by catching the PicklingError in [MRunner.run_tasks()](https://github.com/pydoit/doit/blob/master/doit/runner.py#L417) as follows:
```
diff -r 8602954e6098 doit/runner.py
--- a/doit/runner.py	Mon May 25 15:29:35 2015 +0800
+++ b/doit/runner.py	Fri May 29 11:16:40 2015 +0200
@@ -418,7 +418,10 @@
         # task queue - tasks ready to be dispatched to sub-processes
         job_q = self.Queue()
         self._run_tasks_init(task_dispatcher)
-        proc_list = self._run_start_processes(job_q, result_q)
+        try:
+            proc_list = self._run_start_processes(job_q, result_q)
+        except pickle.PicklingError as exc:
+            raise InvalidTask(repr(exc))
 
         # wait for all processes terminate
         proc_count = len(proc_list)
```

schettino72:
I created a #56, hopefully after that is implemented I will be able to land all the fixes. I am a bit lost at the moment... Your effort won't be wasted, please just have a bit of patience.

schettino72:
@gstorer we finally got a CI running: https://ci.appveyor.com/project/schettino72/doit

I don't how to trigger it for an old pull-request. Can you update your branch to latest master and push? 
(I am hoping appveyor also runs on PRs) After that I will review your patch. Thanks.

schettino72:
So I have another WIP branch https://github.com/pydoit/doit/compare/test

My branch modifies how to pickle MRunner objects only:
 - it excludes the whole TaskDispatcher object, so no need to modify how TaskDispatcher is pickled
 - it also excludes the reporter object, than it pass a reference just the reporter class when using multi-processing.

There is a big problem of test getting stuck on python2.7 (https://ci.appveyor.com/project/schettino72/doit/build/job/gsqqqe94ih2hocms) no idea why.

There is still other failing tests but I suspect they are issues with the tests itself...

I am closing this pull-request, my `test` branch is the way forward. Discussion is going back to original issue #50

PR #55, opened 2015-05-29
Allowing matches of multiple target regexes.

felixfontein:

(If all delayed task loaders provide `'.*'` as a regex, everything will be made. Not great, but it's not so easy to do better.)

### DIFF
--- a/doit/cmd_run.py
+++ b/doit/cmd_run.py
@@ -116,6 +116,18 @@ opt_pdb = {
     }
 
 
+# use ".*" as default regex for delayed tasks without explicitly specified regex
+opt_auto_delayed_regex = {
+    'name': 'auto_delayed_regex',
+    'short': '',
+    'long': 'auto-delayed-regex',
+    'type': bool,
+    'default': False,
+    'help':
+"""Uses the default regex ".*" for every delayed task loader for which no regex was explicitly defined"""
+}
+
+
 class Run(DoitCmdBase):
     doc_purpose = "run tasks"
     doc_usage = "[TASK/TARGET...]"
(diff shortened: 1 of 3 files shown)


schettino72:
Thanks @felixfontein .

What do you think about adding a command line option  to the `run` command, meaning that all DelayedTasks that do not specify a  target_regex would use this `'.*'`.

I guess this way you would get the behaviour you want without having to do any modification in your tasks. You would just need to pass this new parameter in Nikola invocation of `doit`. I guess that would finally be a solution that satisfy both of us :)

felixfontein:
Adding that command line option sounds good. I'll try that later (probably not today, though).

I'm still not really happy with the current implementation, because it tends to build too much (if you don't have good regexes). But at least it will _also_ build the target you specify, so I can live with it :-)

I have some plans on how to improve this, but so far I couldn't get it implemented -- there was always something lurking somewhere in doit conflicting with my ideas ;) But then, we can always improve this later.

felixfontein:
Actually, for Nikola I'll ensure that `'.*'` will be specified as a regexp for every delayed task where no regexp is given -- at least the ones created by Nikola itself. If any plugin decides to use delayed task loaders on its own, they also have to handle regexps on their own...

felixfontein:
Ok, so how shall we proceed from now? :)

schettino72:
Actually I didnt really check your patch in detail yet. Sorry no much time lately.

* Can you make the pull-request to `master` instead of the branch `delayed-target-regex` (of course include the branch)  ?

* It needs unit-tests

* [optional] If if you feel like implementing my suggestion of a new command line option, that can come together or in a separate PR, up to you.

felixfontein:
Ok, 2 (somewhat basic) and 3 are done. I'll create a new PR for the whole branch.

felixfontein:
Continued in #58.

PR #57, opened 2015-06-07
Allows to specify target of a DelayedTask on command line.

### LINKED ISSUE #20: specify target of a DelayedTask on command line

felixfontein:

There's an option to `doit run` called `--auto-delayed-regex` which takes `'.*'` as a regex for every delayed loaded task for which no regex is specified explicitely.

fixes #20

### DIFF
--- a/doit/control.py
+++ b/doit/control.py
@@ -52,9 +52,18 @@ class TaskControl(object):
             self._def_order.append(task.name)
 
         # expand wild-card task-dependencies
+        unpack_all = False
         for task in six.itervalues(self.tasks):
-            for pattern in task.wild_dep:
-                task.task_dep.extend(self._get_wild_tasks(pattern))
+            if len(task.wild_dep) > 0:
+                unpack_all = True
+                break
+        if unpack_all:
+            # execute all delayed loaders to make sure all tasks are actually there
+            while self._unpack_delayed_task():
+                pass
+            for task in six.itervalues(self.tasks):
+                for pattern in task.wild_dep:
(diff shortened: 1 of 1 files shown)


felixfontein:
Hmm, something didn't work when creating this one. I'll try again.

Issue #60, opened 2015-06-12
issues on handling multiple matches delayed target regex


schettino72:
Refers to tasks missing from PR #58 

- [x] need docs
- [ ] when `--auto-delayed-regex` is used, if a target that doesnt exist (or a typo) is passed in the command, it will be silently ignored. It should raise an error saying the target/task/command doesnt exist

felixfontein:
Isn't the second also true when `--auto-delayed-regex` is not used?

schettino72:
> Isn't the second also true when --auto-delayed-regex is not used?

Right, it happens as long as the typo matches a regex. With `--auto-delayed-regex` it always match a regex...

felixfontein:
I'm not sure if you are aware of this, but the current version still builds too much in some situations. Consider the dodo file (slight modification of `doc/tutorial/delayed.py`):
``` .py
import glob

from doit import create_after

@create_after(executed='early', target_regex='.*\.out')
def task_build():
    for inf in glob.glob('*.in'):
        yield {
            'name': inf,
            'actions': ['cp %(dependencies)s %(targets)s'],
            'file_dep': [inf],
            'targets': [inf[:-3] + '.out'],
            'clean': True,
        }

@create_after(executed='build', target_regex='.*')
def task_final():
    return {
        'actions': ['cp a.out a.out.2'],
        'file_dep': ['a.out'],
        'targets': ['a.out.2'],
        'clean': True,
    }

def task_early():
    """a task that create some files..."""
    inter_files = ('a.in', 'b.in', 'c.in')
    return {
        'actions': ['touch %(targets)s'],
        'targets': inter_files,
        'clean': True,
    }
```
will build `a.in`, `a.out`, `b.in`, `b.out`, `c.in` and `c.out` when running `doit run c.out`.

schettino72:
@felixfontein I guess this finally completed. Please check again everything is working as expected.
Thanks for your patience and work on pushing for this feature.

felixfontein:
Cool! Thanks a lot for implementing most parts of this feature! :)
A quick test also showed that it seems to work as expected. Since there weren't big changes since the last branch I've been using extensively, I guess it should be good by now.

felixfontein:
Oh, one thing (completely unrelated!) which I just noticed: if I run `nikola build bla` and it cannot build `bla`, it complains with
```
ERROR: cmd `run` invalid parameter: "bla". Must be a task, or a target.
Type "doit list" to see available tasks
```
[...]

PR #63, opened 2015-06-14
Outputting an error message if no delayed-loaded task built target.

felixfontein:
Addressing second part of #60.

### DIFF
--- a/doit/cmd_run.py
+++ b/doit/cmd_run.py
@@ -249,7 +249,9 @@ class Run(DoitCmdBase):
                 run_args.append(num_process)
 
             runner = RunnerClass(*run_args)
-            return runner.run_all(self.control.task_dispatcher())
+            result = runner.run_all(self.control.task_dispatcher())
+            self.control.post_run()
+            return result
         finally:
             if isinstance(outfile, str):
                 outstream.close()
--- a/doit/control.py
+++ b/doit/control.py
@@ -34,7 +34,7 @@ class TaskControl(object):
         self.tasks = {}
         self.targets = {}
         self.auto_delayed_regex = auto_delayed_regex
-        self.targets_to_search_for = []
+        self.targets_to_search_for = {}
 
         # name of task in order to be executed
         # this the order as in the dodo file. the real execution
@@ -225,7 +225,7 @@ class TaskControl(object):
                        'Type "doit list" to see available tasks')
                 raise InvalidCommand(msg % filter_)
             else:
-                self.targets_to_search_for.append(filter_)
+                self.targets_to_search_for[filter_] = delayed_matched
         return selected_task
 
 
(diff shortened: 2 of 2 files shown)


schettino72:
I was thinking about implementing it in a different way. I would check if any task from a delayed creator has the target right after the task is created. It is more complicated to implemented than what you did but has 2 advantages:

 - you could give a more detailed error message mentioning which delayed creators might create that task.
 - you could cancel the loading/creation of other tasks right after finding the task that creates the specified target.

felixfontein:
What do you mean by more detailed error message? Do you want it to include a list of delayed tasks for which the regex matched? That can easily be added by turning `targets_to_search_for` into a dict, mapping a target to a list of task names which generate it. (The newest commit now does this.)

felixfontein:
The second point you mentioned seems interesting; in fact, I'm hoping that it might help to fix the built-too-much problem mentioned in #60. But I don't see how it helps with the error message.

felixfontein:
Actually, here's a try of mine to implement point 2: https://github.com/felixfontein/doit/commit/4c8cefe2f8ec588a5257d3f3c02b955b9c6911c9
It doesn't help with the problem mentioned in #60, though.

schettino72:
@felixfontein check this patch https://github.com/pydoit/doit/commit/d6d457e2e580501ea3517bae9cbac049f5afba48

schettino72:
I guess we can forget my comment about better error messages. I guess error messages is more important with typos... and in that case it would be better to have a very simple error message.

felixfontein:
Your patch (after removing `set()`) also doesn't help with the problem mentioned in #60, but it looks nicer :)

schettino72:
it fixes #60 for me. are you sure you didnt modify the example? can you post the output you get?

felixfontein:
Ok, I did some more tests. It works *sometimes* for me, and sometimes not. Maybe half of the times, it builds all `.out` files, and half of the times only `c.out`.

If it builds everything, I get:
```
$ doit run c.out
.  early
.  build:b.in
.  build:c.in
.  build:a.in
```

schettino72:
I updated my branch. needed to use OrderedDict...

But note how it depends on the order the tasks were defined. If you define `task_final` before `task_build` it wont work.

TODO:

 - [ ] tests
 - [ ] error message if target not found

felixfontein:
How do you want the error message be implemented? I don't really see how to do it much differently than in https://github.com/felixfontein/doit/commit/a2a40992aa27245458d9dbc929ee2bace2995944.

schettino72:
I guess same message as when not using DelayedTask should be ok.

felixfontein:
My question was more at how outputting the error should be implementing. Is adding a `post_run()` function fine as in this original PR?

schettino72:
> My question was more at how outputting the error should be implementing. Is adding a post_run() function fine as in this original PR?

I would need to implement it to decide :) I would like to avoid having a `post_run()`... But not sure how complicated would be to modify the `RegexGroup` class to track all tasks and targets created.

When all tasks from a RegexGroup have already been created we can check if the targets it is expected to create is on any task, right?

felixfontein:
See #70 for a try to implement that.

PR #64, opened 2015-06-15
modifications to make doit run on python 2.6

JohannesBuchner:
These changes allow doit to run also on python 2.6


schettino72:
Sorry python 2.6 support was dropped a while ago, check the CHANGES file. You will have to keep this branch yourself or use an old version.

PR #70, opened 2015-06-28
Another try on adding an error message when delayed task target was not found.

felixfontein:
As discussed in #63.


schettino72:
merged on master. thanks

PR #72, opened 2015-07-08
Fix apostrophes in zsh Tab completion

Kwpolska:
Fixes getnikola/nikola#1870.

Signed-off-by: Chris Warrick <kwpolska@gmail.com>

### DIFF
--- a/doit/cmd_completion.py
+++ b/doit/cmd_completion.py
@@ -129,15 +129,15 @@ class TabCompletion(DoitCmdBase):
         # '(-c|--continue)'{-c,--continue}'[continue executing tasks...]' \
         # '--db-file[file used to save successful runs]' \
         if opt.short and opt.long:
-            tmpl = ("'(-{0.short}|--{0.long})'{{-{0.short},--{0.long}}}'"
-                    "[{help}]' \\")
+            tmpl = ('"(-{0.short}|--{0.long})"{{-{0.short},--{0.long}}}"'
+                    '[{help}]" \\')
         elif not opt.short and opt.long:
-            tmpl = "'--{0.long}[{help}]' \\"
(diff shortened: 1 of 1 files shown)


doit/cmd_completion.py
```diff
@@ -129,15 +129,15 @@ def _zsh_arg_line(opt):
...
         elif not opt.short and opt.long:
-            tmpl = "'--{0.long}[{help}]' \\"
+            tmpl = '"--{0.long}[{help}]" \\'
         elif opt.short and not opt.long:
-            tmpl = "'-{0.short}[{help}]' \\"
+            tmpl = '"-{0.short}[{help}]" \\'
         else: # without short or long options cant be really used
             return ''
-        ohelp = opt.help.replace(']', r'\]')
+        ohelp = opt.help.replace(']', r'\]').replace("'", r"\'")
```
schettino72: I don't understand... Since you modified the lines above replacing `'` with `"`, I guess here you should escape `"` not `'`.
I guess if you had modified only this line it would fix the problem.

Kwpolska: It was actually supposed to be replacing `"` with `\"`, that’s an error. Fixed and squashed.

Modifying this line only would not fix it, as escapes do not work in single quotes in zsh.

schettino72:
thanks! I didnt actually test it... I am trusting you :)

I guess Nikola includes a generated zsh complention file that needs to be manually updated. That could be done now, no need to wait next doit release.

schettino72:
And of course it breaks the tests... :( I will fix that myself.

Kwpolska:
Nikola does not include a generated file at all, users just run `nikola tabcompletion` and take it from there. I’m not even sure if it’s documented anywhere.

Issue #24, opened 2015-02-09
show reason task is not up-to-date


madsflensted:
Show the reason why a task is (would be) executed, e.g. "Missing target: x.y.z" or "Dependencey updated: a.b.c". 

This should be implemented as part of `doit info` command.

schettino72:
@madsflensted this is now on master. Can you please test and check as you expected. Use:

$ doit info --status task_name

schettino72:
@felixfontein I was thinking that it would be better to display only the status and reason when the option `--status` is used. The normal output with file_dep, targets would be omitted.  What do you think?

felixfontein:
@schettino72: both is fine for me. I only use `doit info` and `doit info --status` for debugging, and I spend most of the time finding out why exactly something changed, so running `doit info` twice instead of once doesn't really matter.

madsflensted:
@felixfontein sorry for the long feedback cycle, summer vacation...
I just pulled the master and gave it a spin, and the extra information is very helpful in debugging my tasks.
Thanks :+1:

felixfontein:
You're welcome! I already had some debugging sessions (before this feature) where I would have really liked something like that, too.

PR #75, opened 2015-07-13
Added functionality to log why a task will be rebuild.

felixfontein:
Addressing #24.

Will show the reason why a task is rebuild on `doit info`.

### DIFF
--- a/doit/cmd_info.py
+++ b/doit/cmd_info.py
@@ -8,7 +8,6 @@ import six
 
 from .cmd_base import DoitCmdBase
 from .exceptions import InvalidCommand
-from .dependency import DependencyException
 
 
 
@@ -24,6 +23,16 @@ def my_safe_repr(obj, context, maxlevels, level):
     return pprint._safe_repr(obj, context, maxlevels, level)
 
 
+opt_show_build_reason = {
+    'name': 'show_build_reason',
+    'short': 'R',
+    'long': 'show-build-reason',
+    'type': bool,
+    'default': False,
+    'help': """Shows reasons why this target will be rebuild. [default: %(default)s]"""
+}
+
+
 class Info(DoitCmdBase):
     """command doit info"""
 
@@ -31,7 +40,9 @@ class Info(DoitCmdBase):
     doc_usage = "TASK"
     doc_description = None
 
-    def _execute(self, pos_args):
+    cmd_options = (opt_show_build_reason, )
+
+    def _execute(self, pos_args, show_build_reason=False):
         if len(pos_args) != 1:
             msg = ('doit info failed, must select *one* task.'
                    '\nCheck `doit help info`.')
@@ -59,20 +70,14 @@ class Info(DoitCmdBase):
                 self.outstream.write('\n{0}:'.format(attr))
                 printer.pprint(getattr(task, attr))
 
-        if self.dep_manager is None:
-            # Avoid failure in tests which don't initialize dep_manager.
-            return
-        rebuild_log = []
-        try:
-            status = self.dep_manager.get_status(task, tasks, rebuild_log=rebuild_log)
-        except DependencyException:
-            # A raised DependencyException means that a file dependency is missing.
-            # Since rebuild_log contains the necessary information, we treat this
-            # the same as a return value of 'run'.
-            status = 'run'
-        if status == 'up-to-date':
-            self.outstream.write('\nIs up to date.\n')
-        else:  # status == 'run'
-            self.outstream.write('\nIs not up to date:\n')
-            for entry in rebuild_log:
-                self.outstream.write('  * {0}\n'.format(entry))
+        if show_build_reason:
+            if self.dep_manager is None:
+                # Avoid failure in tests which don't initialize dep_manager.
+                return
+            status = self.dep_manager.get_status(task, tasks, get_log=True)
+            if status.status == 'up-to-date':
+                self.outstream.write('\nIs up to date.\n')
+            else:  # status.status == 'run' or status.status == 'error'
+                self.outstream.write('\nIs not up to date:\n')
+                for entry in status.rebuild_log:
+                    self.outstream.write('  * {0}\n'.format(entry))
--- a/doit/cmd_info.py
+++ b/doit/cmd_info.py
@@ -79,5 +79,31 @@ class Info(DoitCmdBase):
                 self.outstream.write('\nIs up to date.\n')
             else:  # status.status == 'run' or status.status == 'error'
                 self.outstream.write('\nIs not up to date:\n')
-                for entry in status.rebuild_log:
-                    self.outstream.write('  * {0}\n'.format(entry))
+                if status.uptodate_false:
+                    self.outstream.write(' * The following uptodate objects evaluate to false:\n')
+                    for utd, utd_args, utd_kwargs in status.uptodate_false:
+                        self.outstream.write('    - {} (args={}, kwargs={})\n'.format(utd, utd_args, utd_kwargs))
+                if status.has_no_dependencies:
+                    self.outstream.write(' * The task has no dependencies.\n')
+                if status.missing_target:
+                    self.outstream.write(' * The following targets do not exist:\n')
+                    for target in status.missing_target:
+                        self.outstream.write('    - {}\n'.format(target))
+                if status.file_dep_checker_changed:
+                    self.outstream.write(' * The file_dep checker changed from {0} to {1}.'.format(*status.file_dep_checker_changed))
+                if status.added_file_dep:
+                    self.outstream.write(' * The following file dependencies were added:\n')
+                    for dep in status.added_file_dep:
+                        self.outstream.write('    - {}\n'.format(dep))
+                if status.removed_file_dep:
+                    self.outstream.write(' * The following file dependencies were removed:\n')
+                    for dep in status.removed_file_dep:
+                        self.outstream.write('    - {}\n'.format(dep))
+                if status.missing_file_dep:
+                    self.outstream.write(' * The following file dependencies are missing:\n')
+                    for dep in status.missing_file_dep:
+                        self.outstream.write('    - {}\n'.format(dep))
+                if status.changed_file_dep:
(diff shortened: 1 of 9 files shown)


felixfontein:
The last commit is kind of a hack to get the unit tests working again. They failed because the `CmdInfo` object had `dep_manager` not set. This decreases the coverage (because the new code isn't tested in any unit test).

schettino72:
Thanks for working on this...

- what about adding a command line option? IMO this information should be displayed only when explicitly asked for.

- It would be nice to display **all** reasons a task is not up-to-date, right now it just gives the first reason found.

- It would be nice if the exit value of the command indicates if the task is up-to-date or not. zero for up-to-date, 1 for not up-to-date.

doit/dependency.py
```diff
@@ -528,7 +528,7 @@ def status_is_ignore(self, task):
         return self._get(task.name, "ignore:")
 
     # TODO add option to log this
```
schettino72: it is not a TODO anymore :)

doit/dependency.py
```diff
@@ -528,7 +528,7 @@ def status_is_ignore(self, task):
...
-    def get_status(self, task, tasks_dict):
+    def get_status(self, task, tasks_dict, rebuild_log=None):
```
schettino72: that is so ugly. `rebuild_log` is a return value not a parameter. This is not a public API. I make sure doit has 100% code coverage to give confidence in making this kind of changes.

doit/dependency.py
```diff
@@ -593,17 +597,23 @@ def get_status(self, task, tasks_dict):
...
+            if rebuild_log is not None:
+                rebuild_log.append('Task has no dependencies.')
             return 'run'
 
         # if target file is not there, task is not up to date
         for targ in task.targets:
             if not os.path.exists(targ):
                 task.dep_changed = list(task.file_dep)
+                if rebuild_log is not None:
+                    rebuild_log.append('Target file "{}" does not exist.'.format(targ))
```
schettino72: It would be better to not repeat the text of the message all the time. I mean I prefer a message like

Target file does not exist:
 - target1
 - target2
 - ...

doit/dependency.py
```diff
@@ -625,11 +640,16 @@ def get_status(self, task, tasks_dict):
             try:
                 file_stat = os.stat(dep)
             except OSError:
+                if rebuild_log is not None:
+                    rebuild_log.append('Dependent file "{}" does not exist.'.format(dep))
                 raise DependencyException("Dependent file '{}' does not exist."
```
schettino72: Since rebuild_log will be returned we need to find another solution for this case instead of raising an exception.

schettino72:
Thinking over I think `get_status()` should return a dict (or namedtuple, or custom object) with:

 - result: a string one of `run`, `up-to-date`, `error` (like missing file_dep)
 - missing_file_dep: a list of file_dep names
 - changed_file_dep: list
 - missing_target: list
 - ...

what do you think?

doit/dependency.py
felixfontein: Now that it lists all reasons, definitely. Removed it.

felixfontein:
I now added a result object and moved the result formatting to `cmd_info`. Also all results are displayed, and all that is only done when explicitly requested via `-R` or `--show-build-reason`.

About the exit value: what should the exit value be in case `-R` et al is not specified? Also 0?

tom-varga:
Will this also work along with --dry-run?
I hope so.

felixfontein:
BTW (completely off-topic), when do you plan to release doit 0.29.0?

felixfontein:
@tom-varga: why shouldn't it? It should not affect `doit run` at all.

doit/cmd_info.py
```diff
@@ -23,6 +23,16 @@ def my_safe_repr(obj, context, maxlevels, level):
     return pprint._safe_repr(obj, context, maxlevels, level)
 
 
+opt_show_build_reason = {
+    'name': 'show_build_reason',
+    'short': 'R',
+    'long': 'show-build-reason',
```
schettino72: IMO this name is too long. Maybe just `status` would be ok. The short name `R` stands for what? and why a capital letter?

doit/cmd_info.py
```diff
@@ -23,6 +23,16 @@ def my_safe_repr(obj, context, maxlevels, level):
...
+    'type': bool,
+    'default': False,
+    'help': """Shows reasons why this target will be rebuild. [default: %(default)s]"""
```
schettino72: r/target will be rebuild/task would be executed

doit/cmd_info.py
```diff
@@ -57,3 +69,41 @@ def _execute(self, pos_args):
             if value:
                 self.outstream.write('\n{0}:'.format(attr))
                 printer.pprint(getattr(task, attr))
+
+        if show_build_reason:
+            if self.dep_manager is None:
+                # Avoid failure in tests which don't initialize dep_manager.
```
schettino72: Is this still required?

doit/dependency.py
```diff
@@ -536,13 +555,27 @@ def get_status(self, task, tasks_dict):
...
+                                    values up-to-date, run or error
 
         task.dep_changed (list-strings): file-dependencies that are not
         up-to-date if task not up-to-date because of a target, returned value
         will contain all file-dependencies reagrdless they are up-to-date
         or not.
         """
+        result = DependencyStatus()
+        if get_log:
+            result.uptodate_false = []
```
schettino72: Since you created a class `DependencyStatus` this initialization should be done in `DependencyStatus.__init__()`

doit/dependency.py
```diff
@@ -426,6 +426,26 @@ def get_state(self, dep, current_state):
...
+class DependencyStatus(object):
+    """Result object for Dependency.get_status.
+
+    @ivar status: (str) one of "run", "up-to-date" or "error"
+    """
+
+    def __init__(self):
+        self.status = None
+
+    def _decide(self, status):
```
schettino72: Why this method has an `_` ? It is used outside of the class.

doit/dependency.py
```diff
@@ -426,6 +426,26 @@ def get_state(self, dep, current_state):
...
+        """Sets state of status object if it wasn't set before."""
+        if self.status is None:
+            self.status = status
+
+    def raise_exception(self):
+        """Raises a DependencyException if the state is 'error'."""
+        if self.status == "error":
+            raise DependencyException("Dependent file '{}' does not exist.".format(self._error_reason))
```
schettino72: I mean this exception could be completely removed. Where this exception is used, it could just deal with the return value.

schettino72:
> BTW (completely off-topic), when do you plan to release doit 0.29.0?

Normally I would not do a release now since there is so few changes since last release and there are few issues WIP. Anyway if you really need a release there is no problem to push one on your request ;)

doit/cmd_info.py
felixfontein: I chose `R` for `reason` (I didn't want to use lowecase `r` because I tend to associate that with `recursive`). But `status` sounds good, too, with shortcut `s`.

doit/cmd_info.py
felixfontein: Makes sense. I was thinking too much in Nikola terms :)

doit/dependency.py
felixfontein: Would you prefer `__`?

doit/dependency.py
felixfontein: That's true, but that might be a more complicated transformation, to make sure nothing breaks.

felixfontein:
It would be nice to have a released version with target regex support and earlytask target specifying on command line. Especially to get the tests in getnikola/nikola#1632 working again. But it's not really that urgent, as I'm not sure when that branch can be merged after all; if there will be another 7.x version of Nikola before Nikola 8 is released, it must wait a bit more.

felixfontein:
(Just checked, Nikola 7.6.2 should be released probably this month.)

Oh, and I'll work on the changes in this thread tomorrow, no time left for today :)

doit/cmd_info.py
felixfontein: Changed to `status`/`s`.

doit/cmd_info.py
felixfontein: Changed.

doit/cmd_info.py
felixfontein: Shouldn't be. Removed.

doit/dependency.py
felixfontein: Done.

doit/dependency.py
felixfontein: Ok, I tried that now.

felixfontein:
Any idea why this one test is failing? I cannot see a connection to my changes.

Kwpolska:
Don't expect Nikola v8 to be released soon.

felixfontein:
I added some more tests for the `get_log=True` option of `get_status` as well as for `cmd_info`; now the coverage is back to 100%.

doit/dependency.py
felixfontein: I now made the `decide()` method local to `get_status` (instead of having it in `DependencyStatus`). Now the only drawback is that `get_status` sets `DependencyStatus._error_message` (which is read by `DependencyStatus.get_error_message()`.

felixfontein: r/`_error_message`/`_error_reason`. And never mind, got rid of underscores here by now.

schettino72:
thanks. I already started to work on this. so, no need for you to make more commits...

felixfontein:
Ok. Thanks!

schettino72:
merged, thanks (rebased it so github cant detect it)

Sorry, I was half-way merging and doing changes when you did your last batch of commits.

I did some changes to try to make get_status simpler because it was already a big function,
so I tried to remove as much complexity as possible from it.

schettino72:
the test failed on travis is #7, hopefully we can get rid of it after #51 is merged.

felixfontein:
Ah, ok. I assumed you probably have seen it fail before.

Issue #74, opened 2015-07-10
Failing on callable tasks with keyword only args


Hinidu:
Hello! Thank you for the great tool!
I found that problem when I was trying to use `shutil.copyfile`:
```py
from shutil import copyfile

def task_copyfile():
    return {'actions': [(copyfile, ['foo', 'bar'])]}
```
It fails with the following error message:
```
Traceback (most recent call last):
  File "C:\Python34\lib\site-packages\doit\doit_cmd.py", line 165, in run
    return command.parse_execute(args)
  File "C:\Python34\lib\site-packages\doit\cmd_base.py", line 122, in parse_execute
    return self.execute(params, args)
  File "C:\Python34\lib\site-packages\doit\cmd_base.py", line 405, in execute
    return self._execute(**exec_params)
  File "C:\Python34\lib\site-packages\doit\cmd_run.py", line 239, in _execute
    return runner.run_all(self.control.task_dispatcher())
  File "C:\Python34\lib\site-packages\doit\runner.py", line 238, in run_all
    self.run_tasks(task_dispatcher)
  File "C:\Python34\lib\site-packages\doit\runner.py", line 204, in run_tasks
    catched_excp = self.execute_task(node.task)
  File "C:\Python34\lib\site-packages\doit\runner.py", line 166, in execute_task
    return task.execute(sys.stdout, sys.stderr, self.verbosity)
  File "C:\Python34\lib\site-packages\doit\task.py", line 376, in execute
    action_return = action.execute(task_stdout, task_stderr)
  File "C:\Python34\lib\site-packages\doit\action.py", line 368, in execute
    kwargs = self._prepare_kwargs()
  File "C:\Python34\lib\site-packages\doit\action.py", line 337, in _prepare_kwargs
    self.args, self.kwargs)
  File "C:\Python34\lib\site-packages\doit\action.py", line 44, in _prepare_kwargs
    argspec = inspect.getargspec(func)
  File "C:\Python34\lib\inspect.py", line 936, in getargspec
    raise ValueError("Function has keyword-only arguments or annotations"
ValueError: Function has keyword-only arguments or annotations, use getfullargspec() API which can support them
```

I'm using python 3.4, according to the [documentation](https://docs.python.org/3.4/library/shutil.html#shutil.copyfile) argument `follow_symlinks` was added to `shutil.copyfile` in python 3.3. With python 2.7 code above works without any problem.
[...]

PR #77, opened 2015-07-23
Handle python3 callables with keyword only args and annotations

Hinidu:
Hi, @schettino72!
I've added the usage of `inspect.getfullargs` in python3 so it should handle python callables with keyword only arguments and annotations.
Though I have one issue: I'm pretty new to the python and I don't know any way to hide the code from interpreter if python2 is used, so new tests generate syntax errors with python2 because keyword only arguments were syntactically incorrect before python3. Do you have any suggestions which I could use to solve this issue?
Original issue: #74
PS. Sorry for submitting the patch later than I was promising (life always gets in the way).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,3 +19,4 @@
  * Simon Conseil - contact <at> saimon <dot> org
  * Kostis Anagnostopoulos - ankostis <at> gmail <dot> com
  * Randall Schwager - schwager <at> hsph <dot> harvard <dot> edu
+ * Pavel Platto - hinidu <at> gmail <dot> com
--- a/doit/action.py
+++ b/doit/action.py
@@ -40,10 +40,19 @@ class BaseAction(object):
         if not task:
             return kwargs
 
-        try:
-            argspec = inspect.getargspec(func)
-        except TypeError: # a callable object, not a function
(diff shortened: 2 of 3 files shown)


schettino72:
I guess you can put the code into a string and than [eval](https://docs.python.org/3/library/functions.html?highlight=eval#eval) it. Of course `eval()` it only on python 3.

You should add your name into the `AUTHORS` file :)

Thanks!

Hinidu:
> I guess you can put the code into a string and than eval it.

Thanks! Nice trick :-)

> You should add your name into the AUTHORS file :)

Done :sunglasses:

schettino72:
thanks. merged. (github didnt detect because i squashed your commits).

By the way, I noticed you use Windows (from your traceback). Wanna help make all tests pass on windows?

Hinidu:
> thanks. merged. (github didnt detect because i squashed your commits).

Thank you!

> By the way, I noticed you use Windows (from your traceback). Wanna help make all tests pass on windows?

I use Windows only on the work, so I would be glad to try next week.

schettino72:
If can take a look at my comment on:
https://github.com/pydoit/doit/issues/50#issuecomment-124849157

appveyor gives me no clue why the test got stuck :(

PR #83, opened 2015-08-04
Quick patch for doc about action arguments.

FrankStain:
As requested in https://groups.google.com/d/msg/python-doit/XfnVbvouoiI/bcdhkTuvBAAJ i've made small description about task arguments.

### DIFF
--- a/doc/tasks.rst
+++ b/doc/tasks.rst
@@ -67,6 +67,13 @@ Classes and built-in functions are not allowed.
 as positional and keywords arguments for the callable.
 see `Keyword Arguments <http://docs.python.org/tutorial/controlflow.html#keyword-arguments>`_.
 
+For ``args`` may be accepted tuple or list object.
+
+.. literalinclude:: tutorial/task_args.py
+
+For ``kwargs`` only `dict` object may be accepted.
+
+.. literalinclude:: tutorial/task_kwargs.py
 
 The result of the task is given by the returned value of the
 ``action`` function.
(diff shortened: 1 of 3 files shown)


doc/tasks.rst
```diff
@@ -168,6 +168,25 @@ check :ref:`tools.LongRunning<tools.LongRunning>` as an example.
 
 
 
+action arguments
+-----------------
+
+You always can pass some arguments into your action in case it's a `python-action`.
```
schettino72: dont need repeat this explanation. It is already said on the section above `python-actions`. You only need to add an example that uses kwargs.

doc/tasks.rst
```diff
@@ -168,6 +168,25 @@ check :ref:`tools.LongRunning<tools.LongRunning>` as an example.
...
+
+For ``args`` may be accepted tuple or list object.
+
+.. literalinclude:: tutorial/task_args.py
+
+For ``kwargs`` only `dict` object may be accepted.
+
+.. literalinclude:: tutorial/task_kwargs.py
+
+Also you may be interested in :ref:`params<task_args.arguments>` and :ref:`getargs<dependencies.getargs>` for passing arguments into actions.
```
schettino72: I try to avoid reference concepts that were not explained yet because first time readers wont be able to understand this.

FrankStain:
Next fix according your suggestions.

schettino72:
rebased, squashed and commited. Thanks. 

I removed the first example you added because there is already an example like that. I also moved it a bit below from where you placed.

PR #88, opened 2015-08-07
permit not-catching of stdout

JohannesBuchner:
This patch allows disabling capture of stdout and stderr.

This is important because progressbars (lines that are updated with "\r" or similar) can not be properly shown by passing the input data on line by line.

The technical solution I chose requires the user to set encoding to None to disable capture.

### DIFF
--- a/doit/action.py
+++ b/doit/action.py
@@ -109,7 +109,8 @@ class CmdAction(BaseAction):
     @ivar save_out: (str) name used to save output in `values`
     @ivar shell: use shell to execute command
                  see subprocess.Popen `shell` attribute
-    @ivar encoding (str): encoding of the process output
+    @ivar encoding (str): encoding of the process output.
+                 if set to None, stdout/stderr will not be captured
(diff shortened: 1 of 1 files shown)


schettino72:
First, thanks for working on this. 

Using the `encoding` to control this is not a nice solution...  

Actually I plan to change this capture mode from line based to byte based (and give option for the user to control it). So I am interested in this but it needs a more careful design. It would be nice to have an example ready to run so that can experiment while working on this.

JohannesBuchner:
Here is an example program:
```
import progressbar
import time

print 'calculating ...'
pbar = progressbar.ProgressBar(widgets=[progressbar.Percentage(), 
	progressbar.Counter('%5d'), progressbar.Bar(), progressbar.ETA()])
for obj in pbar(range(10000)):
	time.sleep(0.006)

print 'calculating done.'
```

schettino72:
since this implementation is not suitable I open #94 to track this. feel free to submit another PR.

PR #100, opened 2015-08-28
Please make the completion output deterministic

lamby:
While working on the "reproducible builds" effort [1], I noticed that doit
generates completion that is non-determistic due to the arbitrary nature
of Python dict ordering.

This patch removes this randomness by sorting prior to output. Once
applied, packages using doit (eg. nikola, etc.) can be built reproducibly
in our toolchain.

 [1]: https://wiki.debian.org/ReproducibleBuilds

Signed-off-by: Chris Lamb <chris@chris-lamb.co.uk>


schettino72:
Merged. thanks.

And thanks for working on "reproducible builds".

PR #107, opened 2015-09-20
Python 3.5 testing on Travis CI

Kwpolska:


schettino72:
Thanks for looking into this.

The release of coverage 4.0 breaks running tests on doit. I will fix on the weekend...

schettino72:
fixed in https://github.com/pydoit/doit/commit/18fb7645fa8a25a0199229a17dce982bad34f7eb

Issue #109, opened 2015-09-27
Equivalent paths are considered different (should use os.path.normpath)


DarwinAwardWinner:
I have one task with a (auto-generated) target of "./some_file" and another task with a dependency of "some_file", and doit failed to recognize the dependency between the two tasks, presumably because is uses string equality to determine whether two paths are not equal. I recognize that, in the presence of symlinks, hard links, bind mounts, and other such things, the general problem of determining whether two paths lead to the same file is difficult, but I think at the very least doit should run `os.path.normpath` before comparing two paths for equality.

schettino72:
There is also the issue about detecting if 2 paths are the same file on case-insensitive file systems.

This came up before and I decided it is a `wontfix`. You can easily use `normpath` yourself when required.

I am keeping this open as `doc` issue to make sure this behavior is documented somewhere.

PR #114, opened 2015-11-12
Add support for Path from pathlib.

Hinidu:
Adding support for `Path` in `file_dep` is quite easy. But clearly this is only the part of the problem: if someone (like me) uses `Path` then he want to use them all other the place. It would be quite strange to pass them to `file_dep` as is but convert to `str` before passing in `targets` for example. I think at least `targets` should support them too and perhaps it would be not very bad idea to support them in `CmdAction`.
@schettino72 what do you think about it?

### DIFF
--- a/doc/tutorial/checker.py
+++ b/doc/tutorial/checker.py
@@ -1,3 +1,10 @@
+from pathlib import Path
+
 def task_checker():
     return {'actions': ["pyflakes sample.py"],
             'file_dep': ["sample.py"]}
+
+def task_checker_pathlib():
+    sample_path = Path("sample.py")
+    return {'actions': [["pyflakes", sample_path]],
+            'file_dep': [sample_path]}
--- a/doit/action.py
+++ b/doit/action.py
@@ -214,15 +214,17 @@ class CmdAction(BaseAction):
 
 
     def expand_action(self):
-        """expand action string using task meta informations
-        @returns (string) - expanded string after substitution
+        """Expand action using task meta informations if action is a string.
+        Convert `Path` elements to `str` if action is a list.
+        @returns: string -> expanded string if action is a string
+                  list - string -> expanded list of command elements
         """
         if not self.task:
             return self.action
 
         # cant expand keywords if action is a list of strings
         if isinstance(self.action, list):
-            return self.action
+            return [str(x) for x in self.action]
 
         subs_dict = {'targets' : " ".join(self.task.targets),
                      'dependencies': " ".join(self.task.file_dep)}
--- a/doit/action.py
+++ b/doit/action.py
@@ -4,6 +4,7 @@
 import subprocess, sys
 from io import StringIO
 import inspect
+from pathlib import PurePath
 from threading import Thread
 
 from .exceptions import InvalidTask, TaskFailed, TaskError
@@ -224,7 +225,18 @@ class CmdAction(BaseAction):
 
         # cant expand keywords if action is a list of strings
         if isinstance(self.action, list):
-            return [str(x) for x in self.action]
+            action = []
+            for element in self.action:
+                if isinstance(element, str):
+                    action.append(element)
+                elif isinstance(element, PurePath):
+                    action.append(str(element))
+                else:
+                    msg = ("%s. CmdAction element must be a str " +
+                           "or Path from pathlib. Got '%r' (%s)")
+                    raise InvalidTask(
+                        msg % (self.task.name, element, type(element)))
+            return action
 
         subs_dict = {'targets' : " ".join(self.task.targets),
                      'dependencies': " ".join(self.task.file_dep)}
--- a/tests/test_action.py
+++ b/tests/test_action.py
@@ -4,6 +4,7 @@ import tempfile
 import textwrap
 import locale
 locale # quiet pyflakes
+from pathlib import PurePath, Path
 
 from io import StringIO, BytesIO
 import pytest
@@ -221,6 +222,12 @@ class TestCmdExpandAction(object):
         task = FakeTask([],[],[], {})
         my_action = action.CmdAction(cmd, task)
         assert cmd == my_action.expand_action()
+
+    def test_list_can_contain_path(self):
+        cmd = ["python", PurePath(TEST_PATH), Path("myecho.py")]
+        task = FakeTask([], [], [], {})
+        my_action = action.CmdAction(cmd, task)
+        assert ["python", TEST_PATH, "myecho.py"] == my_action.expand_action()
 
 
 class TestCmd_print_process_output(object):
--- a/tests/test_action.py
+++ b/tests/test_action.py
@@ -229,6 +229,12 @@ class TestCmdExpandAction(object):
         my_action = action.CmdAction(cmd, task)
         assert ["python", TEST_PATH, "myecho.py"] == my_action.expand_action()
 
+    def test_list_should_contain_strings_or_paths(self):
+        cmd = ["python", PurePath(TEST_PATH), 42, Path("myecho.py")]
+        task = FakeTask([], [], [], {})
+        my_action = action.CmdAction(cmd, task)
+        assert pytest.raises(action.InvalidTask, my_action.expand_action)
+
 
 class TestCmd_print_process_output(object):
     def test_non_unicode_string_error_strict(self):
--- a/tests/test_task.py
+++ b/tests/test_task.py
@@ -192,6 +192,13 @@ class TestTaskDeps(object):
         assert [(None, None, None), (True, None, None)] == my_task.uptodate
 
 
+class TestTaskPathlibSupport(object):
+
+    def test_targets_can_be_path(self):
+        my_task = task.Task("Task X", ["taskcmd"],
+                            targets=["123", Path("456"), PurePath("789")])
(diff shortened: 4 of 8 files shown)


schettino72:
Yes. Support it everywhere... 

I was even thinking about actually storing the PosixPath instead of converting it to string.
I would need to try implementing the code to see exactly what would be the consequences of that change...

Hinidu:
> I was even thinking about actually storing the PosixPath instead of converting it to string.
I would need to try implementing the code to see exactly what would be the consequences of that change...

What's the benefit? I can imagine only two implementations:
1) Store in `file_dep` both `str`s and `Path`s and don't forget to cast items in `file_dep` when needed - cumbersome and error-prone.
2) Use `Path`s internally and convert `str`s provided by the user to them - pointless as I can see because I think `doit` doesn't use many operations on the file paths besides path comparison, though rich set of handy operations is the main feature of `pathlib` in my point of view.

But maybe I miss something important.

schettino72:
> What's the benefit?

I was thinking about supporting passing a `glob` as a `file_dep` so the glob could be lazily evaluated. 
But the glob result is not even an instance of PosixPath, it is just a generator. 
So not really much relation to what you are doing...

Another reason would be to have a more robust file comparison. See #109, but I guess that is not worth the problem.

Conclusion: you are right :smile: , just convert to string.

schettino72:
- You need to add a dependency to `pathlib` on `setup.py` because doit still supports python3.3. 

- Note some test failures on windows, check appveyor

Hinidu:
> I was thinking about supporting passing a glob as a file_dep so the glob could be lazily evaluated. 
But the glob result is not even an instance of PosixPath, it is just a generator.
So not really much relation to what you are doing...

It also can be a good thing to do, but I agree that not in this PR.

> Another reason would be to have a more robust file comparison. See #109, but I guess that is not worth the problem.

Hmm, I didn't think about `some_file` and `./some_file`. I thought only about case-insensitive file systems.
Now it looks more benefitial. I will try to investigate the converting of all user strings to `Path`s - maybe it can be done as the next step after accepting `Path`s from the user.

> Conclusion: you are right :smile: , just convert to string.

Yeah, I think it would be a good first step :-)

> You need to add a dependency to pathlib on setup.py because doit still supports python3.3.
Note some test failures on windows, check appveyor

Ok.

Hinidu:
> Note some test failures on windows, check appveyor

It looks like these failed tests didn't run before this PR on Windows. I assumed it by the number of tests:
- PR: `============== 4 failed, 702 passed, 15 skipped in 13.44 seconds ==============`
- master: `=================== 699 passed, 21 skipped in 13.94 seconds ===================`

I have added only one test.
These failed tests have a decorator: `@pytest.mark.skipif("os.system('strace -V') != 0")`
Do you have any ideas why `strace -V` could change its behaviour on AppVeyor?

Hinidu:
@schettino72 I've commited support for `Path` in `targets` and `CmdAction`. Waiting for your comments on implementation. Could you point to the specific parts of the documentation which I should update to reflect these changes?

Hinidu:
Perhaps it would be reasonable to do type checking for `targets` and `CmdAction` similar to existing for `file_dep` to provide more meaningful error message to the user.
Current approach will convert to `str` anything user will provide. I can imagine that some users want from time to time put some `int`s in `CmdAction`, but I think that other types in these places much rather would be a mistake.
What do you think about it?

schettino72:
> Do you have any ideas why strace -V could change its behaviour on AppVeyor?

Sorry. No idea.

> Could you point to the specific parts of the documentation which I should update to reflect these changes?

When it introduces file_dep add an example using pathlib. Then in other places (targets, ...) just mention that it also supports pathlib too.

> Perhaps it would be reasonable to do type checking for targets and CmdAction similar to existing for file_dep to provide more meaningful error message to the user.

go ahead :smile: 

Sorry I didnt review the code yet... and probably wont do that next week also :sweat:

schettino72:
> Do you have any ideas why strace -V could change its behaviour on AppVeyor?

I re-run the tests in an old commit just to double check.
Now AppVeyor provides a strace (from cygwin).
https://ci.appveyor.com/project/schettino72/doit/build/1.0.97/job/qlx1air685fiaxvs#L223

I dont want to debug that... lets just disable the test if strace not available OR it is on windows.
If you dont want to do this change yourself, it is fine... but please create another issue to track this. thanks.

schettino72:
thanks. one more thing. add your name in the AUTHORS file :grinning:

schettino72:
and an entry on CHANGES file

tests/test_task.py
```diff
@@ -150,7 +151,12 @@ def test_dependencyStringIsFile(self):
         my_task = task.Task("Task X", ["taskcmd"], file_dep=["123","456"])
         assert set(["123","456"]) == my_task.file_dep
 
-    def test_file_dep_must_be_string(self):
+    def test_file_dep_can_be_path(self):
+        my_task = task.Task("Task X", ["taskcmd"],
+                            file_dep=["123", Path("456"), PurePath("789")])
+        assert {"123","456", "789"} == my_task.file_dep
+
+    def test_file_dep_must_be_string_or_path(self):
```
schettino72: test_file_dep_path()

tests/test_task.py
```diff
@@ -150,7 +151,12 @@ def test_dependencyStringIsFile(self):
         my_task = task.Task("Task X", ["taskcmd"], file_dep=["123","456"])
         assert set(["123","456"]) == my_task.file_dep
 
-    def test_file_dep_must_be_string(self):
+    def test_file_dep_can_be_path(self):
```
schettino72: test_file_dep_path

tests/test_action.py
```diff
@@ -222,6 +223,12 @@ def test_string_list_cant_be_expanded(self):
         my_action = action.CmdAction(cmd, task)
         assert cmd == my_action.expand_action()
 
+    def test_list_can_contain_path(self):
+        cmd = ["python", PurePath(TEST_PATH), Path("myecho.py")]
+        task = FakeTask([],[],[], {})
```
schettino72: needs a whitespace after every comman

doit/action.py
```diff
@@ -214,15 +214,17 @@ def execute(self, out=None, err=None):
 
 
     def expand_action(self):
-        """expand action string using task meta informations
-        @returns (string) - expanded string after substitution
+        """Expand action using task meta informations if action is a string.
+        Convert `Path` elements to `str` if action is a list.
```
schettino72: better say exactly what you are doing (converting all elements to string), nothing specific about Path.

schettino72:
looks good. thanks. just need docs now.

tests/test_action.py
Hinidu: Fixed

tests/test_task.py
Hinidu: Fixed

tests/test_task.py
Hinidu: Fixed

doit/action.py
Hinidu: I will add validation and raising `InvalidTask` if the user provides something that is nor `str` nor `Path`. I mentioned it earlier: https://github.com/pydoit/doit/pull/114#issuecomment-157982283

Hinidu:
> thanks. one more thing. add your name in the AUTHORS file :grinning:

I'm already there :wink:

I'll add info to `CHANGES` and docs.

Hinidu:
@schettino72 thank you for review.
I've pushed requested changes, type checking in `targets` and `CmdAction` and docs update.
PR looks ready to me now.

schettino72:
thanks. squashed and merged.

There was a typo in the docs, already fixed. 
Next time you can check for typos using `doit spell`.

doc/tutorial/checker.py
```diff
@@ -1,3 +1,10 @@
+from pathlib import Path
+
 def task_checker():
     return {'actions': ["pyflakes sample.py"],
             'file_dep': ["sample.py"]}
+
+def task_checker_pathlib():
```
schettino72: actually I would prefer to have a separate example... to keep examples really simple.
But I merged anyway.

Hinidu:
@schettino72 thank you for merging!
By the way this PR can help partially solve the problem from #109 if user will provide `Path` instances instead of strings. `str(Path('./afile')) == str(Path('afile')) == 'afile'`. But it doesn't resolve symbolic links and `..` though.

doc/tutorial/checker.py
Hinidu: I agree - it is a good thing to concentrate reader on one feature at once. I can remove this example here and add a new section like `pathlib support` with example which uses `pathlib` both in `file_dep`, `targets` and `CmdAction`.

schettino72: PR with whatever improvements are always welcome.

schettino72:
> By the way this PR can help partially solve the problem from #109 if user will provide Path instances instead of strings. str(Path('./afile')) == str(Path('afile')) == 'afile'. But it doesn't resolve symbolic links and .. though.

To tell you the truth I dont care about this feature. I am worried it could have a negative effect on performance. I dont know how I would implement that. If you to give a try go ahead :)

PR #115, opened 2015-11-19
Skip strace tests on windows and cygwin.

Hinidu:
AppVeyor now provides `strace` from `cygwin`. This fact was discovered in https://github.com/pydoit/doit/pull/114#issuecomment-158134093. This PR provides a quick fix for that problem.
Actually I didn't use `doit strace` command, maybe later someone who finds it useful will fix it on Windows :-)


schettino72:
thanks

PR #129, opened 2016-03-11
Prevent PythonAction from modifying task attributes by passing copies

rbeagrie:
Currently, if a PythonAction calls pop() or otherwise modifies dependencies, targets or changed, these changes are reflected in the task itself. This leads to the dependency or target in question not being added to the doit database properly after execution has finished.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,7 +10,7 @@
  * Michael Gliwinski - https://launchpad.net/~tzeentch-gm
  * Vadim Fint - mocksoul <at> gmail <dot> com
  * Thomas Kluyver - https://bitbucket.org/takluyver
- * rbeagrie - https://bitbucket.org/rbeagrie
+ * Rob Beagrie - http://rob.beagrie.com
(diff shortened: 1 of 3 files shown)


schettino72:
Actually PythonAction being able to modify task attributes is a feature :grinning: 

It is even documented: https://github.com/pydoit/doit/blame/master/doc/tasks.rst#L205

We should add unit tests for that...

schettino72:
It is a quite powerful feature.. if you search the mailing list you can find some use cases for that.

rbeagrie:
Ah sorry I missed that in the docs! It would only take me 10 minutes to invert the unit tests I wrote and test the correct behaviour, if you think that would be helpful?

schettino72:
>  It would only take me 10 minutes to invert the unit tests I wrote and test the correct behaviour, if you think that would be helpful?

that would be cool. thanks

I was also thinking about a simple example. Like creating a target where the filename includes the timestamp. It is a simple example but show how this feature can be powerful...

PR #131, opened 2016-03-22
Fix race condition in doit.tools.create_folder

Hinidu:
Hi again!
While making possible to run build of my project concurrently with `-n NUMBER` option I discovered that `create_folder` has a race condition: if many tasks try to create the same folder then it is possible that between checking directory existence and actual creation of the directory in the next line another task will create that directory.
[Since python 3.2](https://docs.python.org/3.4/library/os.html#os.makedirs) `os.makedirs` has nice `exist_ok` argument that will do the same thing but [without such problems](https://hg.python.org/cpython/file/tip/Lib/os.py#l233).

### DIFF
--- a/doit/tools.py
+++ b/doit/tools.py
@@ -15,8 +15,7 @@ result_dep # pyflakes
 # action
 def create_folder(dir_path):
     """create a folder in the given path if it doesnt exist yet."""
-    if not (os.path.exists(dir_path) and os.path.isdir(dir_path)):
-        os.makedirs(dir_path)
+    os.makedirs(dir_path, exist_ok=True)
 
 
 # title


schettino72:
Dropping python 2 is great :grin: 

I am extremely busy these days. I will merge as soon as I get some extra time, yeah i could just click a button but I dont like github merging stuff for me...

Hinidu:
Of course! Do it at any time :wink:

schettino72:
merged: https://github.com/pydoit/doit/commit/169a1df1e82b0e7832e082b274b28242701cf659

I amended the commit to add an entry on CHANGES. so github thinks this was not merged...

thanks :smile:

Hinidu:
Thank you!

PR #132, opened 2016-04-03
Added unit tests to check actions can modify targets etc.

rbeagrie:
Supercedes #129 

New PR created because I borked the old one somehow...


schettino72:
thanks

PR #135, opened 2016-05-03
Fix auto command on OS X

smutch:
This PR starts the fsevents observer on the main thread, allowing the auto command to pickup file changes and run the appropriate tasks.  This may address #49 (and possibly #101) by fixing the current implementation without moving to using Watchdog.

The filewatcher test fails on OS X both before and after this patch, but testing the auto command manually after applying the patch seems to show that the auto command works.  I am unaware if there are any unintended consequences of running the observer on the main thread though as I have no previous experience with fsevents.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -21,3 +21,4 @@
  * Randall Schwager - schwager <at> hsph <dot> harvard <dot> edu
  * Pavel Platto - hinidu <at> gmail <dot> com
  * Gerlad Storer - https://github.com/gstorer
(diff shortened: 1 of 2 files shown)


schettino72:
thanks. I will merge this soon. probably this week.

Please add your name to the authors file :smile:

smutch:
Great!  Glad its useful.

Thanks for a great piece of software.  I only started using a week ago, but it is already proving really us for my research!

schettino72:
thanks

amitu:
When is the next release planned?

PR #142, opened 2016-07-02
Fix KeyError on doit list --status when missing file dependency

poliquin:
Tasks missing a file dependency will cause a `KeyError` on `doit list --status` because `List.STATUS_MAP` lacks an `'error'` key.

This pull requests resolves the issue by showing `'E'` next to the task name.

Example of a problematic `dodo.py` that causes a `KeyError` on `doit list --status` before this fix:

``` python
def task_a():
    """Task A"""
    return {
        'actions': ['echo "hello" > a.txt'],
        'file_dep': ['b.txt']  # assume b.txt does not exist
    }
```

### DIFF
--- a/doit/cmd_list.py
+++ b/doit/cmd_list.py
@@ -65,7 +65,7 @@ class List(DoitCmdBase):
                    opt_list_private, opt_list_dependencies, opt_template)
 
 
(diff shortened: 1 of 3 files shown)


schettino72:
thanks. squashed and merge. sorry for delay.

You should add yourself to the `AUTHORS` file

schettino72:
really merged (and pushed) this time. sorry.

PR #147, opened 2016-08-15
Fix #144

### LINKED ISSUE #144: "reset_dep" fails when a task is missing both dependencies and targets
Example dodo.py file:

```python
def task_a():
    return {'actions':['echo "Tiger, tiger, burning bright," > %(targets)s'],
            'targets':['line_1.txt'],
            'uptodate':[True]}

def task_b():
    return {'actions':['cat %(dependencies)s > %(targets)s',
                       'echo "In the forests of the night," >> %(targets)s'],
            'file_dep':['line_1.txt'],
            'targets':['line_2.txt']}
def task_c():
    return {'actions':['cat %(dependencies)s > %(targets)s',
                       'echo "What immortal hand or eye," >> %(targets)s'],
            'file_dep':['line_2.txt'],
            'targets':['line_3.txt']}

def task_d():
    return {'actions':['cat %(dependencies)s > %(targets)s',
                       'echo "Could frame thy fearful symmetry?" >> %(targets)s'],
            'file_dep':['line_3.txt'],
            'targets':['line_4.txt']}
```

Running `doit reset-dep` when line_3.txt is missing works as expected:

```
$ doit && rm .doit.db line_3.txt && echo "resetting deps:" && doit reset-dep
-- a
-- b
.  c
.  d
resetting deps:
skip a
processed b
processed c
failed d (Dependent file 'line_3.txt' does not exist.)
```

It also works when line_4.txt is missing:

```
$ doit && rm .doit.db line_4.txt && echo "resetting deps:" && doit reset-dep
-- a
-- b
-- c
.  d
resetting deps:
skip a
processed b
processed c
processed d
```

However, the command throws an exception if both line_3.txt and line_4.txt are missing (i.e. both the dependency and the target):

```
$ doit && rm .doit.db line_[34].txt && echo "resetting deps:" && doit reset-dep
-- a
-- b
-- c
.  d
resetting deps:
skip a
processed b
processed c
  File "/usr/lib/python2.7/site-packages/doit/doit_cmd.py", line 168, in run
  File "/usr/lib/python2.7/site-packages/doit/cmd_base.py", line 122, in parse_execute
  File "/usr/lib/python2.7/site-packages/doit/cmd_base.py", line 405, in execute
  File "/usr/lib/python2.7/site-packages/doit/cmd_resetdep.py", line 68, in _execute
    self.dep_manager.save_success(task, result_hash=result)
  File "/usr/lib/python2.7/site-packages/doit/dependency.py", line 535, in save_success
    state = self.checker.get_state(dep, self._get(task.name, dep))
  File "/usr/lib/python2.7/site-packages/doit/dependency.py", line 417, in get_state
    timestamp = os.path.getmtime(dep)
  File "/usr/lib/python2.7/genericpath.py", line 62, in getmtime
    return os.stat(filename).st_mtime
OSError: [Errno 2] No such file or directory: 'line_3.txt'
```

rbeagrie:
Fix for issue #144 

Catch the exception thrown when dependencies and targets are both missing. Also added regression test for this issue.

### DIFF
--- a/doit/cmd_resetdep.py
+++ b/doit/cmd_resetdep.py
@@ -1,5 +1,6 @@
 from .cmd_base import DoitCmdBase, check_tasks_exist
 from .cmd_base import subtasks_iter
+import os
 
 
 class ResetDep(DoitCmdBase):
@@ -45,17 +46,23 @@ to a task that has already run.
 
         write = self.outstream.write
         for task in task_list:
+            print(task)
             # Get these now because dep_manager.get_status will remove the task
             # from the db if the checker changed.
             values = self.dep_manager.get_values(task.name)
             result = self.dep_manager.get_result(task.name)
 
-            res = self.dep_manager.get_status(task, tasks)
-            if res.status == 'error':
-                # Skip exception when a depencency file is missing, and force
-                # the state computation
-                write("failed {} ({})\n".format(task.name, res.get_error_message()))
+            missing_deps = []
+            for dependency in task.file_dep:
+                if not os.path.exists(dependency):
+                    missing_deps.append(dependency)
+
+            if len(missing_deps) > 0:
+                write("failed {} (Dependent file '{}' does not "
+                "exist.)\n".format(task.name, "', '".join(missing_deps)))
                 continue
+
+            res = self.dep_manager.get_status(task, tasks)
 
             # An 'up-to-date' status means that it is useless to recompute the
             # state: file deps and targets exists, the state has not changed,
@@ -66,11 +73,7 @@ to a task that has already run.
 
             task.values = values
 
-            try:
-                self.dep_manager.save_success(task, result_hash=result)
-            except OSError as e:
-                write("failed {} ({})\n".format(task.name, e))
-                continue
+            self.dep_manager.save_success(task, result_hash=result)
 
             write("processed {}\n".format(task.name))
 
--- a/doit/cmd_resetdep.py
+++ b/doit/cmd_resetdep.py
@@ -46,7 +46,6 @@ to a task that has already run.
 
         write = self.outstream.write
         for task in task_list:
-            print(task)
             # Get these now because dep_manager.get_status will remove the task
             # from the db if the checker changed.
             values = self.dep_manager.get_values(task.name)
@@ -57,12 +56,12 @@ to a task that has already run.
                 if not os.path.exists(dependency):
                     missing_deps.append(dependency)
 
+            res = self.dep_manager.get_status(task, tasks)
+
             if len(missing_deps) > 0:
                 write("failed {} (Dependent file '{}' does not "
                 "exist.)\n".format(task.name, "', '".join(missing_deps)))
                 continue
-
-            res = self.dep_manager.get_status(task, tasks)
 
             # An 'up-to-date' status means that it is useless to recompute the
             # state: file deps and targets exists, the state has not changed,
--- a/doit/cmd_resetdep.py
+++ b/doit/cmd_resetdep.py
@@ -56,12 +56,12 @@ to a task that has already run.
                 if not os.path.exists(dependency):
                     missing_deps.append(dependency)
 
-            res = self.dep_manager.get_status(task, tasks)
-
             if len(missing_deps) > 0:
                 write("failed {} (Dependent file '{}' does not "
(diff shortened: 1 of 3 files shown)


doit/cmd_resetdep.py
```diff
@@ -65,7 +65,13 @@ def _execute(self, pos_args=None):
                 continue
 
             task.values = values
-            self.dep_manager.save_success(task, result_hash=result)
+
+            try:
+                self.dep_manager.save_success(task, result_hash=result)
+            except OSError as e:
+                write("failed {} ({})\n".format(task.name, e))
```
schettino72: This the Error message will be different than the your use-case when you removed only `line_3.txt`. 

I would prefer if we explicitly check all `file_deps` exist. So we wont even need to check for `res.status == "error"`.

tests/test_cmd_resetdep.py
```diff
@@ -81,6 +81,33 @@ def test_missing_file_dep(self, depfile):
...
+                      targets=['tests/data/dependency4'])
+
+        output = StringIO()
+        tasks = [task_a, task_b, task_c]
+        cmd_reset = CmdFactory(ResetDep, outstream=output, task_list=tasks,
+                               dep_manager=depfile)
+        cmd_reset._execute()
+
+        got = [line.strip().split()[:2] for line in output.getvalue().split('\n') if line]
+        got_status = [g[0] for g in got]
```
schettino72: i dont like much logic in tests. It is preferable to hard-code things and repeat yourself.

tests/test_cmd_resetdep.py
```diff
@@ -81,6 +81,33 @@ def test_missing_file_dep(self, depfile):
...
+        got_task_names = [g[1] for g in got]
+        expected_task_names = [t.name for t in tasks]
+
+        assert 'processed' in got_status
+        assert 'failed' in got_status
+        assert sorted(got_task_names) == sorted(expected_task_names)
```
schettino72: same here. not logic. just hard-code stuff. tests should be as simple as possible. no need normal code good practices...

rbeagrie:
Sorry about the commit spam - tests are passing on my local PC but not on Appveyor so I'm having trouble debugging. I'm still not sure what the problem is, one test is complaining that the database is in use by another process...?

schettino72:
no worries about spam :grin: 

Looks like the error is not caused by you. just ignore it.

schettino72:
i just re-run an old commit (that was successful), it now fails. So the problem must be some change on appveyor...

Of course you are welcome to fix the issue. But better do this in another PR.

rbeagrie:
OK great! Let me know if you're OK with this PR and I'll squash the commits.

doit/cmd_resetdep.py
```diff
@@ -65,7 +71,9 @@ def _execute(self, pos_args=None):
                 continue
 
             task.values = values
+
```
schettino72: i like a lot of blank lines but this seems excessive.

tests/conftest.py
```diff
@@ -31,6 +31,22 @@ def remove_dependency():
 
 # fixture to create a sample file to be used as file_dep
 @pytest.fixture
+def dependency2(request):
```
schettino72: Copy and paste here is not acceptable. 
Note that removing copy and paste here will NOT increase the complexity in the test itself (this code is on conftest).
Hopefully it will not sound contradictory or arbitrary comparing to my previous review.

doit/cmd_resetdep.py
```diff
@@ -50,13 +51,18 @@ def _execute(self, pos_args=None):
             values = self.dep_manager.get_values(task.name)
             result = self.dep_manager.get_result(task.name)
 
-            res = self.dep_manager.get_status(task, tasks)
-            if res.status == 'error':
-                # Skip exception when a depencency file is missing, and force
-                # the state computation
-                write("failed {} ({})\n".format(task.name, res.get_error_message()))
+            missing_deps = []
```
schettino72: i guess this is a good candidate for a single line list-comprehension.

rbeagrie:
I'm guessing that this sort of factory approach is the kind of thing you were thinking about for generating multiple similar fixtures? I couldn't find much documentation about how to do this with pytest so not sure if I'm missing a more obvious solution here...

schettino72:
merged thanks. sorry for delay. i forgot this one.

PR #166, opened 2017-02-16
Fix #161 -- use extras_require to specify per-platform/per-Python dependencies

### LINKED ISSUE #161: Crashes on Python 3.3 due to missing pathlib
I tried switching Nikola’s doit requirement to `doit==0.30.0` for Python 3, but [it fails](https://travis-ci.org/getnikola/nikola/jobs/187164264) on Python 3.3 because `pathlib` is needed. This happens because pip does not notice the extra dependency. The correct way to do this is with [environment markers](https://www.python.org/dev/peps/pep-0496/). We use them in Nikola for doit (in requirements.txt):

```
doit>=0.28.0,<=0.29.0 ; python_version <= '3.3'
doit>=0.28.0 ; python_version >= '3.4'
```

That can also be used for platform checks (pyinotify/macfsevents).

Kwpolska:
This fixes #161. pip understands `extras_require` and uses them properly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ addons:
     packages:
       - strace
 
+before_install:
+  - pip install -U pip setuptools
+
 install:
   - pip install .
   - pip install -r dev_requirements.txt python-coveralls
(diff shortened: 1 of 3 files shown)


schettino72:
failed. "darwin" is not valid on 3.3 and 3.4.

PR #175, opened 2017-04-22
Fix passing environment variables to CmdAction

### LINKED ISSUE #171: Passing environment variables to CmdAction is broken
It seems that [d58a73](https://github.com/pydoit/doit/commit/d58a73aaf6b01d7474a4923bba02fc02f0298547#diff-093aedc4845e9217a7220d8739946096R192) broke the ability to pass environment variables into `CmdAction`: if I now pass the `env` kwarg into the `CmdAction` constructor, this leads to a traceback:

`  File "/Users/onno/anaconda/envs/p3/lib/python3.5/site-packages/doit/action.py", line 204, in execute
    **self.pkwargs)
TypeError: type object got multiple values for keyword argument 'env'`

A minimal `dodo.py` to reproduce:

```python
from doit.action import CmdAction

def task_env_call():
    return {'actions': [CmdAction('echo Hi', env={'ENV_VAR', 'test'})]}
```

I guess that a workaround might be to prepend the environment variables to be set to the shell command string.

onnodb:
This should fix #171. It also includes a small test for passing environment variables to subprocesses.

### DIFF
--- a/tests/sample_process.py
+++ b/tests/sample_process.py
@@ -1,11 +1,17 @@
 #! /usr/bin/env python
 
 # tests on CmdTask will use this script as an external process.
-# 3 or more arguments. process return error exit (166)
-# arguments "please fail". process return fail exit (11)
-# first argument is sent to stdout
-# second argument is sent to stderr
+# - If you call this script with 3 or more arguments, the process returns
+#   exit code (166).
+# - If you call this script with arguments "please fail", it returns exit
+#   code (11).
+# - If you call this script with arguments "check env", it verifies the
+#   existence of an environment variable called "GELKIPWDUZLOVSXE", with
+#   value "1". If the variable is not found, the process returns exit code (99).
+# - Otherwise, any first argument gets written to STDOUT. Any second argument
+#   gets written to STDERR.
 
+import os
 import sys
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


schettino72:
thanks. please give me sometime to merge this as I will not be available next few days.

onnodb:
Sure, no rush. Thanks for your time and efforts in maintaining pydoit :-)

tests/sample_process.py
```diff
@@ -17,6 +23,12 @@
         sys.stdout.write("out ouch")
         sys.stderr.write("err output on failure")
         sys.exit(11)
+    # check env
+    if len(sys.argv) == 3 and sys.argv[1]=='check' and sys.argv[2]=='env':
+        if os.environ.get('GELKIPWDUZLOVSXE') == '1':
```
schettino72: so what `GELKIPWDUZLOVSXE` means? :grin:

schettino72:
thanks. merged.

Are you on AUTHORS file? if not please add your name there.

tests/sample_process.py
onnodb: Just a random string really. Although I'm sure we can come up with a backronym 😁

onnodb:
Thanks, but such a tiny pull request is something I wouldn't like to call 'authoring' ;-) Once I make a more significant contribution, I'd be happy to add myself to the list.

Issue #173, opened 2017-04-20
Documentation for private/hidden tasks is ambiguous


sean-abbott:
Without experimentation, it's not clear if private hidden tasks with a name that starts with underscore mean that you should do `_task_private_task()` or `task__private_task()` or something else entirely.

sean-abbott:
In case anyone is curious, you can either use `task__private_task()` or in the return dictionary, you can have a basename key, `'basename': '_private_task'`

schettino72:
well, it is no ambiguous to me. The name of task is what comes after `task_`.

Anyway it would be nice to have an example in the docs...
Would you please submit a pull-request to the docs.

PR #178, opened 2017-05-09
add code and output samples for private tasks

sean-abbott:
Adds a little more flavor documentation to the private tasks explanations.

### DIFF
--- a/doc/tasks.rst
+++ b/doc/tasks.rst
@@ -500,6 +500,27 @@ private/hidden tasks
 
 If task name starts with an underscore '_', it will not be included in the output.
 
+.. literalinclude:: tutorial/private_task.py
+
+.. code_block:: console
+
+    $ doit
+    .  hello
+    $ doit dumpdb
+    DBM type is 'dbm.gnu'
+    b'_private_get_username' -> {   '_values_:': {},
+        'checker:': 'MD5Checker',
+        'deps:': [],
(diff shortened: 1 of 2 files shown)


sean-abbott:
addresses #173

doc/tasks.rst
```diff
@@ -500,6 +500,27 @@ private/hidden tasks
...
 
+.. literalinclude:: tutorial/private_task.py
+
+.. code_block:: console
+
+    $ doit
+    .  hello
+    $ doit dumpdb
+    DBM type is 'dbm.gnu'
+    b'_private_get_username' -> {   '_values_:': {},
```
schettino72: oh. what an ugly output with those byte types `b'xxx'`

doc/tasks.rst
```diff
@@ -500,6 +500,27 @@ private/hidden tasks
 
 If task name starts with an underscore '_', it will not be included in the output.
 
+.. literalinclude:: tutorial/private_task.py
+
+.. code_block:: console
+
+    $ doit
+    .  hello
+    $ doit dumpdb
```
schettino72: I do not think this is a good example. The output is very lengthy and the `dumpdb`  is an advanced command that most users would not know it exist or its purpose.

Better would be to have a private task that prints out something into the console. So you could see that it was executed although its name does not show up. 

what do you think?

schettino72:
closing this as it seems inactive...

PR #186, opened 2017-08-04
Support env variable for seek file

balkierode:
Supports setting an environment variable `DOIT_SEEK_FILE` instead of passing `--seek-file` in commandline


schettino72:
thanks. i implemented this in a different way...

PR #189, opened 2017-08-24
Use `executable` from `sys` module instead of hardcoded `python`

frenzymadness:
Hello.

I met this issue when trying to run tests of doit module on a Python3-only system. Problem is that on a lot of systems `python` still means Python 2 and when you have only Python 3 installed, there is no `python` command at all.

I think that the best will be to execute commands inside test suite under the same Python as test suite itself.

This fix will also enable to run the test suite with various Python versions.

What do you think?

### DIFF
--- a/tests/test___main__.py
+++ b/tests/test___main__.py
@@ -1,5 +1,7 @@
 import subprocess
+from sys import executable
+
 
 def test_execute(depfile_name):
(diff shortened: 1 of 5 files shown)


schettino72:
nice. I didnt know this `sys.executable` existed :grin:

PR #192, opened 2017-09-04
Use load_ipython_extension to register doit magic functions

tonyfast:
With this update

```ipython
%load_ext doit.tools
```
replaces

```ipython
from doit.tools import register_doit_as_IPython_magic
register_doit_as_IPython_magic()
```
This works using [`load_ipython_extension`](http://ipython.readthedocs.io/en/stable/config/extensions/#writing-extensions) instead of `register_doit_as_IPython_magic` .

[Here is an example notebook with this change.](http://nbviewer.jupyter.org/gist/tonyfast/46dcf89845ec319288c59aebd5fa75bc)

### DIFF
--- a/doc/tools.rst
+++ b/doc/tools.rst
@@ -78,8 +78,7 @@ First you need to register the new magic function into ipython shell.
 
 .. code-block:: pycon
 
-    >>> from doit.tools import register_doit_as_IPython_magic
-    >>> register_doit_as_IPython_magic()
+    >>> %load_ext doit.tools
 
 
 .. Tip::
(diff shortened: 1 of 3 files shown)


schettino72:
@ankostis , since you wrote the original code can you help review this?

Thanks @tonyfast , it looks better but a few questions:

- is this `load_ipython_extension` a new thing? is there any advantage in using the register style?
- can we have `%load_ext doit` instead of `%load_ext doit.tools`
- it would be nice to maintain compatibility for a while (add a deprecation warning), or just support both styles forever.
- please add an entry on CHANGES and AUTHORS

tonyfast:
* is this load_ipython_extension a new thing? is there any advantage in using the register style?

   > `load_ipython_extension` has been a thing for as long as I have known; it provides extra functionality to the register style. [_This documentation highlights writing extensions._](http://ipython.readthedocs.io/en/stable/config/extensions/#writing-extensions)

* [can we have %load_ext doit instead of %load_ext doit.tools](https://github.com/pydoit/doit/pull/192/commits/f0630bda2eb9cc4a9b75addb1518417bca1ac311)

* [it would be nice to maintain compatibility for a while (add a deprecation warning), or just support both styles forever.](https://github.com/pydoit/doit/pull/192/commits/642eb9158a042cc9c9550c38c1e16e7f473acf79)

schettino72:
have you noticed your changes broke the build?

tonyfast:
Missed that.  I sorted the pyflakes error and now everything passes.

schettino72:
thanks. I have finally (squashed) and merged this. I tested it and it works fine :)

Seems github is not smart enough to detect I merged manually.

PR #219, opened 2017-11-03
fixed MD5 test to also work if test file has been cloned with crlf endings

moltob:
While preparing my development environment for the clean/forget command, I stumbled upon a failing test on Windows, Python 3.6.3, resulting from the fact that my GIT was checking out with crlf, leading to a different MD5 hash value.

I fixed the test by allowing both, the lf and crlf based hashes to be accepted. I verified the added hash in http://onlinemd5.com/.

It's weird that appveyor does not catch this, it seems to clone with lf only... Is that possible?


schettino72:
Thanks. 
Not really happy with checking the result against 2 values... hope i never have to change this :grimacing:

Issue #218, opened 2017-11-03
clean should optionally include forget


moltob:
The current implementation of the ``clean`` command does not include ``forget``, i.e. if no other dependencies are outdated, doit assumes tasks are still up-do-date and effective, even after ``clean`` was processed.

While this may be the useful in some situations, it is a bit counter intuitive for users not being aware of the semantic difference of the two commands. In particular in a scenario, where doit-based tasks are rolled out to a software development team, they often will expect ``clean`` to invalidate all cleaned tasks, as they are used to from other build engines like make, ninja, and so on.

As suggested by @schettino72 it would be helpful to extend the current ``clean`` command such that it optionally also forgets tasks. The new behavior can be requested per command line option to ``clean`` and should also become a configuration option to make it the default ``clean`` behavior within the context of a project.

I will look at the existing ``clean`` and ``forget`` commands and try to implement it in my fork.

PR #220, opened 2017-11-05
added option to forget during clean

moltob:
Clean can now also forget about previous task execution be either adding the `--hard` option or by specifying `clean_hard=True` during doit configuration, see updated docs and unit tests.

This implements the feature request in #218.

### DIFF
--- a/doit/cmd_clean.py
+++ b/doit/cmd_clean.py
@@ -31,7 +31,7 @@ opt_clean_cleanall = {
 
 opt_clean_forget = {
     'name': 'cleanforget',
-    'long': 'clean-forget',
+    'long': 'forget',
     'type': bool,
     'default': False,
     'help': 'also forget tasks after cleaning',
--- a/doit/cmd_clean.py
+++ b/doit/cmd_clean.py
@@ -57,8 +57,7 @@ class Clean(DoitCmdBase):
                 if forget_tasks:
                     self.dep_manager.remove(task.name)
 
-        if forget_tasks:
-            self.dep_manager.close()
+        self.dep_manager.close()
 
     def _execute(self, dryrun, cleandep, cleanall, hard, pos_args=None, clean_hard=False):
         """Clean tasks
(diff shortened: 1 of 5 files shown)


doit/cmd_clean.py
```diff
@@ -29,32 +29,46 @@
...
         cleaned = set()
+        forget_tasks = hard and not dryrun
         for task in tasks:
             if task.name not in cleaned:
                 cleaned.add(task.name)
                 task.clean(self.outstream, dryrun)
+                if forget_tasks:
+                    self.dep_manager.remove(task.name)
 
+        if forget_tasks:
```
schettino72: i guess we should always call `close()`. No need the `if-clause`.

doit/cmd_clean.py
```diff
@@ -29,32 +29,46 @@
...
+            self.dep_manager.close()
 
-    def _execute(self, dryrun, cleandep, cleanall, pos_args=None):
+    def _execute(self, dryrun, cleandep, cleanall, hard, pos_args=None, clean_hard=False):
```
schettino72: You are not supposed to add 2 parameters. Why 2?

schettino72: The param with `name` as` hard` should be able to use it from both command-line and config. But as I suggested change it to `forget`. A `clean_XXX` param on `_execute` is not necessary.

moltob:
The reason for the second new arg was to prevent a nonspeaking doit config parameter like `hard` or the new name `forget`. Having that in `DOIT_CONFIG` is not intuitive. You still want it that way or am I missing sth?

schettino72:
Not sure what you mean by 'nonspeaking'. you mean someone reading that wont know what is it for ? 
config from INI fornat does not have this problem because you can set a config value specifically for a comand.

maybe DOIT_CONFIG should support values for specific commands too...
meanwhile if that bothers you just use the name 'clean_forget' only .

moltob:
Perfect, I'll do just that.

schettino72:
you should also add your name to AUTHORS file :smile:

moltob:
I'm through with the adaptations, please pick up the review.

schettino72:
thanks. merged.

i did some small fixes.

PR #225, opened 2018-01-08
Add a Redis backend

saimn:
I use this since a few months and it allows me to run multiple doit processes in parallel on different machines (with a shared NFS folder). I guess it should possible to run the test on Travis, I can try if you like the idea.

### DIFF
--- a/doit/cmd_base.py
+++ b/doit/cmd_base.py
@@ -5,7 +5,7 @@ from collections import deque
 from . import version
 from .cmdparse import CmdOption, CmdParse
 from .exceptions import InvalidCommand, InvalidDodoFile
-from .dependency import CHECKERS, DbmDB, JsonDB, SqliteDB, Dependency
+from .dependency import CHECKERS, DbmDB, JsonDB, SqliteDB, Dependency, RedisDB
 from .plugin import PluginDict
 from . import loader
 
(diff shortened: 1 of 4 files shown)


schettino72:
This is pretty cool. I think lots of people (including myself) will like using this. Thanks.

But no way redis can be a dependency for doit. 
I would like to see this as a doit plugin [1].
What do you think?

Testing code needs to be improved a bit so we dont need to check for DB type.

[1] http://pydoit.org/extending.html#plugins

saimn:
redis could be just an optional dependency. I was not aware that it was possible to use the plugin system for backends, so yeah maybe it can be a plugin. Given the limitations of the other backends and the fact that redis a well-know solution for this, I think it deserves to be in core, but no strong feelings.

schettino72:
> Given the limitations of the other backends and the fact that redis a well-know solution for this, I think it deserves to be in core, but no strong feelings.

I agree it is a limitation for **some** use-cases.

I know setting up a full project for some small plugins is very annoying. I dont want to put much burden on people contributing plugins, but on the other hand I dont have enough bandwidth to maintain even core `doit`... and I keep very high standard on doit that needs tests and reviews. In this point a plugin gives more freedom to developers.

`redis` is something I would definitely want as an "official" plugin. So we can use the docs and maybe even the same GIT repository. Would that be a good compromise? 

ps. sorry for delay, but I will be able to (maybe) really do some work again on doit in february.

saimn:
Yes having some plugins in core can be a compromise, as you prefer (I understand the maintenance burden, no worries).

I looked quickly and it seems that there are packages to mock redis for tests:
https://github.com/jamesls/fakeredis
https://github.com/ClearcodeHQ/pytest-redis
https://github.com/locationlabs/mockredis
This will make it easier to test, the most difficult part is probably choosing one of the packages ;)

saimn:
https://github.com/saimn/doit-redis

Issue #130, opened 2016-03-22
Prohibit concurrent execution of tasks using shared resource


Hinidu:
Hi, Eduardo!
I want to use doit concurrent capabilities in my project but I have a problem:
Some of my tasks use Photoshop and I can run only one of them simultaneously :disappointed: 
Do you have any suggestions which I can use to enable concurrent builds but prohibit running of many such tasks?
Thank you in advance!

Hinidu:
After creating this issue I've got an idea: I can set task_dep for each task using Photoshop equal to the previous such task. It seems like it should work. I will try it and tell you how it will work.

Hinidu:
I've implemented described approach and it works quite well. Perhaps we can discuss it and include something like this in doit itself? Or describe it in documentation? Are you interested?
That's my current implementation
```python
from collections import defaultdict

_group_last_task_names = defaultdict(lambda: None)

def put_in_serialized_group(group_name, task_name_base, task):
    last_task_name = _group_last_task_names[group_name]
    if last_task_name is not None:
        if 'task_dep' in task:
            task['task_dep'].append(last_task_name)
        else:
            task['task_dep'] = [last_task_name]

    if task_name_base is None:
        task_name = task['name']
    else:
        task_name = '%s:%s' % (task_name_base, task['name'])
    _group_last_task_names[group_name] = task_name

    return task
```
I'm not very satisfied by this name but I couldn't figure out a better one (`serialized` is like serialized transactions in the database :fearful:).

Usage looks like:
```python
yield put_in_serialized_group('photoshop', 'task_base_name', {
    # actual task properties
})
```

Hinidu:
Yesterday I forgot that except `name` tasks can have `basename`. This version I believe should work correctly for all kinds of tasks:
```python
from collections import defaultdict

_group_last_task_names = defaultdict(lambda: None)

def put_in_serialized_group(group_name, task_basename, task):
    last_task_name = _group_last_task_names[group_name]
    if last_task_name is not None:
        if 'task_dep' in task:
            task['task_dep'].append(last_task_name)
        else:
            task['task_dep'] = [last_task_name]

    if task_basename is None:
        if 'basename' in task:
            task_name = task['basename']
        else:
            raise ValueError('No basename provided')
    elif 'name' in task:
        task_name = '%s:%s' % (task_basename, task['name'])
    else:
        task_name = task_basename
    _group_last_task_names[group_name] = task_name

    return task
```

schettino72:
[...]

PR #226, opened 2018-01-08
Add option to delay the start of processes

saimn:
Because it is sometimes useful to avoid having all the processes doing the same thing (I/O, OpenMP blocks) at the same time.

### DIFF
--- a/doc/tasks.rst
+++ b/doc/tasks.rst
@@ -596,6 +596,32 @@ You can control the verbosity by:
 * from command line, see :ref:`verbosity option<verbosity_option>`.
 
 
+delayed execution
+-----------------
+
+For :ref:`parallel execution <parallel-execution>` it can be useful to delay
+the start of the first task executed in a process. This can be achieved with
+the ``delay`` option:
+
+.. literalinclude:: tutorial/subtasks-delay.py
+
+.. code-block:: console
+
+    $ doit run -v 2 -n 2
+    task file0.txt started (H:M:S+57)
+    .  create_file:file0.txt
+    .  create_file:file1.txt
+    .  create_file:file2.txt
+    task file1.txt started (H:M:S+59)
+    task file2.txt started (H:M:S+59)
+    .  create_file:file3.txt
+    .  create_file:file4.txt
+    task file3.txt started (H:M:S+59)
+    task file4.txt started (H:M:S+59)
+    .  create_file:file5.txt
+    task file5.txt started (H:M:S+59)
+
+
 pathlib
 --------
 
--- a/doit/runner.py
+++ b/doit/runner.py
@@ -5,6 +5,7 @@ from multiprocessing import Process, Queue as MQueue
 from threading import Thread
 import pickle
 import queue
+import time
 
 import cloudpickle
 
@@ -411,8 +412,15 @@ class MRunner(Runner):
         # ### END DEBUG
 
         proc_list = []
-        for _ in range(self.num_process):
+        for i in range(self.num_process):
             next_job = self.get_next_job(None)
+            try:
+                delay = next_job.task_dict.get('delay') or 0
+            except AttributeError:
+                pass
+            else:
+                if i > 0 and delay:
+                    time.sleep(delay)
             if next_job is None:
                 break # do not start more processes than tasks
             job_q.put(next_job)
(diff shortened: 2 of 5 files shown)


schettino72:
Needs docs and tests.

I would like to see some generic task-parameters. So that non-doit core can make use of custom parameters without having to modify doit core code.

tests/test_runner.py
```diff
@@ -681,6 +682,23 @@ def test_all_processes(self, reporter, monkeypatch, dep_manager):
...
+        run = runner.MRunner(dep_manager, reporter, num_process=2)
+        run._run_tasks_init(td)
+        result_q = Queue()
+        task_q = Queue()
+
+        t0 = time()
+        proc_list = run._run_start_processes(task_q, result_q)
+        run.finish()
+        tottime = time() - t0
+        assert tottime > 1
```
saimn: It not that easy to test, so I choose to simply test that the total time is greater than the delay.

doc/tasks.rst
```diff
@@ -596,6 +596,32 @@ You can control the verbosity by:
...
+    .  create_file:file1.txt
+    .  create_file:file2.txt
+    task file1.txt started (H:M:S+59)
+    task file2.txt started (H:M:S+59)
+    .  create_file:file3.txt
+    .  create_file:file4.txt
+    task file3.txt started (H:M:S+59)
+    task file4.txt started (H:M:S+59)
+    .  create_file:file5.txt
+    task file5.txt started (H:M:S+59)
```
saimn: Not sure what's the best place to put this section. Also the example is meant to show that only the first task of a process is delayed.

saimn:
And Travis fails with the same strace errors as the other PR, so probably related to some Travis change.

doc/tasks.rst
```diff
@@ -596,6 +596,32 @@ You can control the verbosity by:
...
 
 
+delayed execution
+-----------------
+
+For :ref:`parallel execution <parallel-execution>` it can be useful to delay
+the start of the first task executed in a process. This can be achieved with
+the ``delay`` option:
+
+.. literalinclude:: tutorial/subtasks-delay.py
```
schettino72: you forgot to commit this file.

doc/tasks.rst
```diff
@@ -596,6 +596,32 @@ You can control the verbosity by:
 * from command line, see :ref:`verbosity option<verbosity_option>`.
 
 
+delayed execution
```
schettino72: doit already uses the name `delayed` to refer to tasks that are not create before execution starts. Not sure this naming will be confusing... can you think about any other name? `lag`?

doit/runner.py
```diff
@@ -411,8 +412,15 @@ def _run_start_processes(self, job_q, result_q):
         # ### END DEBUG
 
         proc_list = []
-        for _ in range(self.num_process):
+        for i in range(self.num_process):
             next_job = self.get_next_job(None)
+            try:
+                delay = next_job.task_dict.get('delay') or 0
```
schettino72: in which occasion the attribute would not be present. tests?

anyway a `get('delay', 0)` would make the try/catch block unnecessary.

doc/tasks.rst
saimn: oops, it must be on my other laptop, I hope I can find it back :confused:

saimn:
> I am not really convinced about this change...

Why ? It's quite simple and really useful when starting processes that all do their I/Os at the beginning to avoid congestion (in my case I load many huge files from the network, and having many parallel reads is really a bottleneck).

> I think this would be more suitable as a command line parameter than as a task parameter.

It depends on tasks. I have tasks (subtasks) that load several Gb files and need a bigger delay, while other tasks may be fine with a smaller or even no delay.

schettino72:
Example: you have a bunch of tasks `A` that are configured with delay `0`, tasks `B` with delay `5` and tasks `C` with delay `1`. 
You execute them `A` will come first, all process will start and in practice you wont have any delay at all. This implementation will not solve the problem...

Have you seen #130? Maybe it can solve your problem in a better way (not counting it is much harder to implement than your solution).

schettino72:
Thanks for adding the file, but you did not answer my questions.

Unfortunately I will not accept this patch. As I pointed above there are some design problems that limit its usability, like it only works on certain situations...

I am open to discuss other ways to solve your problem. Sorry it took me a few rounds of review to finally understand and decide this wont be a good addition to `doit`

PR #228, opened 2018-01-08
Sync changes in the db after each task execution.

saimn:
I no more need this since I use the Redis backend (#225) but it is something you should consider adding (at least with an option if you think it would be too costly for short tasks).
Thanks to this the task execution status is no more lost if doit crashes before the end, and it allows to monitor the progress while doit is still running (I will open another issue about this!).

### DIFF
--- a/doit/dependency.py
+++ b/doit/dependency.py
@@ -84,6 +84,8 @@ class JsonDB(object):
         finally:
             db_file.close()
 
+    sync = dump
+
     def set(self, task_id, dependency, value):
         """Store value in the DB."""
         if task_id not in self._db:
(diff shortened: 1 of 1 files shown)


schettino72:
I think we should just recommend people to use Redis if they want this because performance really sucks. I will leave this PR open so I remember to update the docs.

Or if anyone thinks this useful I could re-consider my decision...

saimn:
It would be useful (even necessary) for something like #229, otherwise you cannot get the up-to-date status of tasks while doit is still running. I didn't measure the cost of this, but tasks that takes "some" time the cost quickly becomes negligible I guess. So the issue is really for short tasks.

schettino72:
ok. lets include this.

saimn:
If someone wants to take over from this, it could be useful for the [report](https://github.com/saimn/doit-report) plugin.

PR #232, opened 2018-01-17
There is no recursive watching of subfolders.

okin:
The documentation should mention that no recursive watching of folders happens.


schettino72:
thanks.

Issue #140, opened 2016-06-15
Print the result of a sh command twice when the command exits abnormally.


congeal:
I call a shell command inside a task to decide pass or fail for the result that another task produces.
The shell command exits with nonzero exit code and prints some failure message if the decision is failure.
Doit prints the failure message that it captured when the task returns with the nonzero exit code, which leads to show the message TWICE when I set the verbosity level '2.'
I think that Doit should not print the captured message when the task failed and its verbosity level is higher than zero.

schettino72:
fair enough. this is an issue since the first days of doit. you are the first to complain :smile: 

do you want give a try providing a patch? :grin:

schettino72:
So sometimes I like the fact the error message is printed twice, once at real time and once again in the end. So lets add an option (non-default) that keeps the current behaviour. But by default if the output is already displayed in realtime it wont be repeated.

PR #238, opened 2018-02-07
fix double error messages from commands #140

imuli:
The old behavior of verbosity level 2 is preserved in verbosity level 3.


schettino72:
The issues says quite clear that a new command line argument should be added. This way you can not control if stdout should be re-displayed or not.
Also to fix 100% the issue you need to take into account the effective verbosity being used by the task. not only the "global" verbosity.

PR #244, opened 2018-02-28
Added Background shell command action

davidbrochart:


schettino72:
Thanks. but for such a trivial implementation I dont think it is worth adding this to doit core.

PR #246, opened 2018-03-01
Guard against using get_var() on uninitialized _CMDLINE_VARS 

### LINKED ISSUE #164: 'get_var' fails while the multiprocess execution if any task uses delayed creation.
@schettino72 , hey! :)

So, i found strange behaviour while trying to implement some stuff.

```.py
#! /usr/bin/doit -f
# -*- coding: utf-8 -*-

from doit import get_var
from doit import create_after

get_var( 'A', None )


def task_foo() :
  return {
    'actions': [ 'echo foo' ],
    'task_dep': [ 'bar' ],
  }
  
@create_after( executed = 'baz' )
def task_bar() :
  for i in range( 10 ) :
    yield {
      'name': 'bar_{}'.format( i ),
      'actions': [ 'echo bar_{}'.format( i ) ]
    }
    
def task_baz() :
  for i in range( 10 ) :
    yield {
      'name': 'baz_{}'.format( i ),
      'actions': [ 'echo baz_{}'.format( i ) ]
    }
```

This code works fine while being executed in single-process mode. But it fails in multi process mode.

```
/root>GetVar.py -n 2
  File "<string>", line 1, in <module>
  File "/py3.4/lib/multiprocessing/spawn.py", line 106, in spawn_main
    exitcode = _main(fd)
  File "/py3.4/lib/multiprocessing/spawn.py", line 116, in _main
    self = pickle.load(from_parent)
  File "/root/GetVar.py", line 7, in <module>
    get_var( 'A', None )
  File "/py3.4/lib/site-packages/doit/doit_cmd.py", line 35, in get_var
    return _CMDLINE_VARS.get(name, default)
AttributeError: 'NoneType' object has no attribute 'get'
  File "<string>", line 1, in <module>
  File "/py3.4/lib/multiprocessing/spawn.py", line 106, in spawn_main
    exitcode = _main(fd)
  File "/py3.4/lib/multiprocessing/spawn.py", line 116, in _main
    self = pickle.load(from_parent)
  File "/root/GetVar.py", line 7, in <module>
    get_var( 'A', None )
  File "/py3.4/lib/site-packages/doit/doit_cmd.py", line 35, in get_var
    return _CMDLINE_VARS.get(name, default)
AttributeError: 'NoneType' object has no attribute 'get'
```

But if i remove the `create_after` decorator, all became works fine even in multi process mode. So, i found it ambiguous. I had seen no notifications in docs about such behaviour.

Can you confirm this is the doit error or not?
Currently i use the last (0.30.0) version of doit.

madig:
This fixes #164.

This is a Windows-specific issue. For an explanation, see https://github.com/pydoit/doit/issues/164#issuecomment-369608965.

### DIFF
--- a/doit/doit_cmd.py
+++ b/doit/doit_cmd.py
@@ -32,6 +32,8 @@ def reset_vars():
     _CMDLINE_VARS = {}
 
 def get_var(name, default=None):
+    if _CMDLINE_VARS is None:
+        return None
     return _CMDLINE_VARS.get(name, default)
 
 def set_var(name, value):
--- a/tests/test_doit_cmd.py
+++ b/tests/test_doit_cmd.py
@@ -45,6 +45,17 @@ class TestRun(object):
         cmd_main(['x=1', 'y=abc'])
         assert '1' == doit_cmd.get_var('x')
         assert 'abc' == doit_cmd.get_var('y')
+        assert None is doit_cmd.get_var('z')
+
+    def test_cmdline_novars(self, monkeypatch):
+        mock_run = Mock()
+        monkeypatch.setattr(Run, "execute", mock_run)
+        cmd_main(['x=1'])
+
+        # Simulate the variable below not being initialized by a subprocess on
+        # Windows. See https://github.com/pydoit/doit/issues/164.
+        doit_cmd._CMDLINE_VARS = None
+        assert None is doit_cmd.get_var('x')
 
     def test_cmdline_vars_not_opts(self, monkeypatch):
         mock_run = Mock()


schettino72:
Uhhmmm. I was looking for a test that would pass on linux and fails on windows without the patch. So at least I could test this using appveyor without relying on asking a windows user to test for me.

Anyway, nice to have but not really required. 

It would be nice if you also add an entry on CHANGES file and add yourself in the list of AUTHORS.

I will merge this when i have some time.

madig:
Haha, yeah, the test is a bit cheeky ;) I'd have to research how to get a MRunner from within a test, will update the PR then.

PR #256, opened 2018-05-03
Raise exception on trying to compute MD5 sum of directory.

madig:
For clarity.

### DIFF
--- a/doit/dependency.py
+++ b/doit/dependency.py
@@ -36,6 +36,9 @@ def get_file_md5(path):
     @param path: (string) file path
     @return: (string) md5
     """
+    if os.path.isdir(path):
+        raise ValueError("Cannot compute MD5 sum of directory. You must "
+                         "specify a file as a dependency instead.")
     with open(path, 'rb') as file_data:
         md5 = hashlib.md5()
         block_size = 128 * md5.block_size
--- a/tests/test_dependency.py
+++ b/tests/test_dependency.py
@@ -34,6 +34,9 @@ def test_md5():
     expected_crlf = "cf7b48b2fec3b581b135f7c9a1f7ae04"
     assert get_file_md5(filePath) in {expected_lf, expected_crlf}
 
+    with pytest.raises(ValueError):
+        get_file_md5(os.path.dirname(filePath))
+
 
 def test_sqlite_import():
     """


doit/dependency.py
```diff
@@ -36,6 +36,9 @@ def get_file_md5(path):
     @param path: (string) file path
     @return: (string) md5
     """
+    if os.path.isdir(path):
```
schettino72: actually this is probably not a good location for doing this check. `check_modified` was carefully implemented to avoid getting `file_stat` more than once. since this is a relatively expensive operation.

madig:
Before, I'd get this stack trace:
```
  File "c:\program files\python36\lib\site-packages\doit\doit_cmd.py", line 177, in run
  File "c:\program files\python36\lib\site-packages\doit\cmd_base.py", line 127, in parse_execute
  File "c:\program files\python36\lib\site-packages\doit\cmd_base.py", line 420, in execute
  File "c:\program files\python36\lib\site-packages\doit\cmd_run.py", line 260, in _execute
    return runner.run_all(self.control.task_dispatcher())
  File "c:\program files\python36\lib\site-packages\doit\runner.py", line 256, in run_all
    self.run_tasks(task_dispatcher)
  File "c:\program files\python36\lib\site-packages\doit\runner.py", line 223, in run_tasks
    self.process_task_result(node, catched_excp)
  File "c:\program files\python36\lib\site-packages\doit\runner.py", line 187, in process_task_result
    self.dep_manager.save_success(task)
  File "c:\program files\python36\lib\site-packages\doit\dependency.py", line 512, in save_success
    state = self.checker.get_state(dep, self._get(task.name, dep))
  File "c:\program files\python36\lib\site-packages\doit\dependency.py", line 400, in get_state
    md5 = get_file_md5(dep)
  File "c:\program files\python36\lib\site-packages\doit\dependency.py", line 39, in get_file_md5
    with open(path, 'rb') as file_data:
PermissionError: [Errno 13] Permission denied: 'sources\\newFont-Bold.ufo'
```
The use case here is that in the font world, the UFO format is a directory with all data for one font contained within in separate files and subdirectories. I originally wanted to pass in the path to the  directory in the hope that doit will know when something within the directory changed.

So where would that check be better kept?

schettino72:
on `get_status()` there is already some code to check if file exist or not. I guess it would make sense to check if file is a directory on same location.

Some use-cases have thousands of file_dep, so performance need to be taken into account.
It would be preferable to not actually not check if file is directory and deal with a raised exception (if possible). Otherwise use the `stat` [1] result. Not the high-level `isdir()` function.

[1] https://docs.python.org/3/library/stat.html#stat.S_ISDIR

madig:
> Some use-cases have thousands of file_dep

Yep, that's our use-case ;)

> It would be preferable to not actually not check if file is directory and deal with a raised exception

You mean in `get_status()`?

schettino72:
@madig so what the status on this?

About checking path vs catching exception. I see you get a permission denied exception and the goal was to give a clear error message... need to be careful not to say a folder is not supported when permission is the real problem.

I think all code changes should go in `get_status()`

madig:
Didn't have time yet, still on my todo list.

schettino72:
Closing this as there is no activity for a long time... feel free to reopen if fix the pending issues.

PR #258, opened 2018-06-07
Command line options -f and -k may be positioned before command. Fixes #204.

### LINKED ISSUE #204: Inconsistent expected position of command wide command line options (`-k`, `-f`)
`doit` has couple of subcommands with respective options.

On the other hand, there are options which are common to all of them, especially when related to location of `dodo.py` file.

Trying to use `dodo.py` located in parent directory (with no `dodo.py` in current one):

```bash
$ doit -k
.  sphinx
Running Sphinx v1.6.4
...
build succeeded.

$ doit -f ../dodo.py 
.  sphinx
Running Sphinx v1.6.4
...
build succeeded.
```
This works fine.

However, trying to call other commands (such as `list`) starts complaining:
```bash
$ doit -f ../dodo.py list
ERROR: Invalid parameter: "list". Must be a command, task, or a target.
Type "doit help" to see available commands.
Type "doit list" to see available tasks.

$ doit -k list
ERROR: Invalid parameter: "list". Must be a command, task, or a target.
Type "doit help" to see available commands.
Type "doit list" to see available tasks.
```

If an option is used after subcommand, all runs well:

```bash
$ doit list -f ../dodo.py
spell    spell checker for doc files
sphinx   build sphinx docs

$ doit list -k
spell    spell checker for doc files
sphinx   build sphinx docs
```

Expected behaviour:

Whichever option may be considered global (namely `--seek-file/-k` and `--file/-f` related to which `dodo.py` file to use) shall be allowed before subcommand is called.

This way it is easy to repeate calls to different subcommands and edit only final part of command line.

barpod:
This PR fixes #204.

### DIFF
--- a/doit/doit_cmd.py
+++ b/doit/doit_cmd.py
@@ -157,6 +157,16 @@ class DoitMain(object):
         # get "global vars" from cmd-line
         args = self.process_args(cmd_args)
 
+        # move -k and -f args to the and of args list so cmd name can be 
+        # used after them
+        if len(args) > 0 and (args[0] == "-k" or args[0] == "--seek-file"):
+            args.append(args.pop(0))
+        if len(args) > 0 and (args[0].startswith("--file=")):
+            args.append(args.pop(0))
+        if len(args) > 1 and (args[0] == "-f"):
+            args.append(args.pop(0))
+            args.append(args.pop(0))
+
         # get specified sub-command or use default='run'
         if len(args) == 0 or args[0] not in sub_cmds:
             specified_run = False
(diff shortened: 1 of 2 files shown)


barpod:
AppVeyor build fails because of https://github.com/pydoit/doit/blob/088d2ba528d65d4cf265035928b64a33eaaa1e63/tests/test_dependency.py#L69. I can reproduce this on my local machine. Is it a common issue?

schettino72:
@barpod just a clarification as you are claiming this was "fixed" before by you.

First your solution breaks very basic use cases like.

$ doit -k list task_name
$ doit -f my_dodo.py task_name

Apart from that your implementation has very poor code quality with copy and paste of defined options.

My solution works not only for hard-coded parameters but for any option parameter that comes that is defined by the task loader.

Next time if you want to contribute it is recommended that you discuss what an accepted solution would be.

PR #265, opened 2018-08-31
[exceptions] Save caught exception as base_exc attribute.

rbdixon:
Reports may wish to make a decision based on the nature of the
underlying exception. In addition to the traceback the CatchedException
also retains the underlying exception as the base_exc attribute.

This is useful for custom tooling that is based on doit. If the task has already displayed a comprehensive error message to the user the Python stack trace is confusing. My custom reporter looks at the base exception and will suppress the stack trace as needed.

### DIFF
--- a/doit/exceptions.py
+++ b/doit/exceptions.py
@@ -49,6 +49,7 @@ class CatchedException(object):
     def __init__(self, msg, exception=None):
         self.message = msg
         self.traceback = ''
+        self.base_exc = exception
 
         if isinstance(exception, CatchedException):
             self.traceback = exception.traceback


rbdixon:
The build failures look to be something unrelated to this one-line change.

schettino72:
> In addition to the traceback the CatchedException also retains the underlying exception as the base_exc attribute.

I think this might be a problem. As far as I remember `CatchedException` is pickled and passed when using a parallel multi-process runner. So if the original Exception contains and object that can not be pickled this would break the runner...  

Could you please try running some sample with multi-process? 
I did not double-check this is really a problem and I could be wrong.
Note that if a base exception is pickleable you would not be able to guarantee that every Exception is pickleable.

> This is useful for custom tooling that is based on doit. If the task has already displayed a comprehensive error message to the user the Python stack trace is confusing. My custom reporter looks at the base exception and will suppress the stack trace as needed.

A task might also return a plain `False` value to indicate that it has failed.
So (maybe) another option for your use-case would be just for you to wrap this logic in the task's action itself without relying on changes in the reporter.

So I am not sure about accepting this PR...
I would be ok on adding more attributes to CachedException as long as they are pickleable.

rbdixon:
ok, thanks. I have not tracked the multi-process work so I was not aware of this.

PR #267, opened 2018-09-03
Add option --sort to 'list' command

### LINKED ISSUE #207: Use natural order of task names when listed
```bash
$ doit list
buildhtml  Build HTML documentation.
buildpdf   Build PDF documentation.
publish    Publish documentation to FTP.
spell	   Spell document source files.
```
Tasks are today listed in alphabetical order.

As seen, it can look quite confusing and user would appreciate natural order:

```bash
$ doit list
spell	   Spell document source files.
buildhtml  Build HTML documentation.
buildpdf   Build PDF documentation.
publish    Publish documentation to FTP.
```

I guess, changing order of listed items is not likely to break anything and I would prefer do it this way by default. Author has good control of order of tasks definitions in dodo file.

If alphabetical order would be (sometime) required, we could add a key "listorder" into `DOIT_CONFIG`.

I had similar issue with Click https://github.com/pallets/click/issues/513 (and I managed to resolve it).

wmvanvliet:
The `--sort` option controls the order in which the tasks are listed.
Currently, two options are supported:

`name`: sort the tasks by name (the default)
`definition`: lists the tasks in the order in which they are defined

closes #207

### DIFF
--- a/tests/test_cmd_list.py
+++ b/tests/test_cmd_list.py
@@ -163,3 +163,33 @@ class TestCmdList(object):
         cmd_list._execute()
         got = [line.strip() for line in output.getvalue().split('\n') if line]
         assert 't做' == got[0]
+
+    def testSort(self):
+        # by default, the task list is ordered by name
+        task_list = list(tasks_sample())
+        output = StringIO()
+        cmd_list = CmdFactory(List, outstream=output, task_list=task_list)
+        cmd_list._execute()
+        got = [line.strip() for line in output.getvalue().split('\n') if line]
+        expected = ['g1', 't1', 't2', 't3']
+        assert expected == got
+
+    def testSortByName(self):
+        # test explicitly sorting task list by name
+        task_list = list(tasks_sample())
+        output = StringIO()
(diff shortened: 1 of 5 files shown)


tests/test_cmd_list.py
```diff
@@ -163,3 +163,33 @@ def test_unicode_name(self, dep_manager):
...
+        # by default, the task list is ordered by name
+        task_list = list(tasks_sample())
+        output = StringIO()
+        cmd_list = CmdFactory(List, outstream=output, task_list=task_list)
+        cmd_list._execute()
+        got = [line.strip() for line in output.getvalue().split('\n') if line]
+        expected = ['g1', 't1', 't2', 't3']
+        assert expected == got
+
+    def testSortByName(self):
```
schettino72: not necessary to repeat the test.

schettino72:
Oh. and it seems docs also need to be updated `cmd-other.rst`. Currently it has the wording:

```
Tasks are listed in alphabetical order, not by order of execution.
```

Instead it should mention both way are possible.

schettino72:
thanks

PR #272, opened 2018-12-10
Make it possible to run config_changed with an custom encoder.

okin:
The new config_changed is great but I had a case where there was need to use a custom encoder.
This PR adds the possibility to specify a special encoder.


doc/uptodate.rst
```diff
@@ -95,6 +95,9 @@ changed. Config values can be a string or dict.
 For dict's the values are converted to string (using `json.dumps()` with `sort_key=True`)
 and only a digest/checksum of the dictionaries keys and values are saved.
 
+If converting the values of the dict requires a special encoder this can be
+passed with the argument ``encoder=...``.
```
schettino72: Can you explicitly mention that `encoder` has the same meaning as used in `json.dumps`.

doit/tools.py
```diff
@@ -49,15 +49,16 @@ class config_changed(object):
     """check if passed config was modified
     @var config (str) or (dict)
     """
-    def __init__(self, config):
+    def __init__(self, config, encoder=None):
```
schettino72: here too need to document parameter. just mention it is the same as in `json.dumps()`

PR #279, opened 2019-01-15
Adding Atomwise success story to `doit` repo

drkeoni:
I've started using `doit` at Atomwise, and it's working very well for various
data processing pipelines.


schettino72:
@drkeoni I am glad `doit` works for you guys. 
Never heard about Atomwise before but looks like pretty cool stuff :+1: 
Thanks for taking the time to write this, it is really appreciated.

Would also be interested to hear some "negative" feedback or what you think doit is lacking. 
I guess I have not got any bug or feature request from you guys...

PR #280, opened 2019-01-17
 Add support for new-style command formatting.

Janfel:
This PR will make it possible to use Python [Format String Syntax][fsyntax] in `cmd-actions` for implicit keyword substitution and `getargs`.

**This change has a slight chance of breaking existing scripts!**

If an existing `dodo.py` file uses `{}` or `{something}` in a `cmd-action`, chances are, the formatter will throw an error.
Existing old-style-syntax will continue to work as expected.

[fsyntax]:https://docs.python.org/3/library/string.html#format-string-syntax

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -26,3 +26,4 @@
  * Mike Pagel - https://github.com/moltob
  * Marijn van Vliet - https://github.com/wmvanvliet
  * Niko Wenselowski - https://github.com/okin
+ * Jan Felix Langenbach - o <dot> hase3 <at> gmail <dot> com
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,7 @@ Changes
  - Fix #261: help organize command options in sections, and improve formatting.
  - Fix #267: `doit list` now has a `--sort` parameter to determine the order in which the tasks are listed.
  - Make it possible to use a custom encoder when using config_changed with a dict.
+ - Make it possible to change the formatter for `cmd-actions` using the configuration value `action_string_formatting`.
 
 
 0.31.1 (*2018-03-18*)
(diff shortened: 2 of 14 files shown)


Janfel:
Are there tests for the old-style syntax i can copy?

schettino72:
also a bit worried about backward compatibility.
What do you think about adding a config like `DOIT_CONFIG` `new_style_string_formatting` ?

Janfel:
I locked the feature behind the flag `new_style_string_formatting`, as you suggested. I also updated the parts of the documentation that seemed relevant and wrote Unit Tests for the feature being and being not explicitly enabled.

schettino72:
@Janfel thanks, you are so fast... 

Thinking over, I guess it would be better to provide a way to migrate to new style and not force people to forever quote their string to both old and new-style format.

So what you think about having: `action_string_formatting` with possible values of `old` (default), `both`, `new`. Then on future releases we could migrate the default from `old` -> `both` -> `new`.

I guess this would be a very small change from your current patch and provide extra value. what do you think?

By the way, also add your name into  AUTHORS file, and add an entry on CHANGES.

Janfel:
This should be everything.

schettino72:
Hi, I am reviewing it and making some changes... WIP here: https://github.com/pydoit/doit/tree/Janfel-janfel-patch-newstyle-formatting

- the unit-tests are broken on windows. see appveyor... the issue is on different escaping on different shells. Thats a can of worm... i would just ignore escaping on tests
- your tests are very comprehensive but also complex. I usually avoid having and logic (IF, FOR) on tests. Also tests should be specific to formatting. no need to test other related features again.
- the formatting should be global, not per task. That just make the implementation easier then what you have done.
- I still want to review the documentation. it needs to make sense when read in order.

I probably wont have more time this week, so please be patient :grin:

schettino72:
squashed and merged. thanks.

PR #284, opened 2019-03-12
Minor modifications to the tutorial, mainly english aesthetics

facundofc:
Lots are trivial things, but even though English is not my native language, I humbly beleive it's a bit better now.

### DIFF
--- a/doc/tutorial_1.rst
+++ b/doc/tutorial_1.rst
@@ -3,12 +3,12 @@
 tutorial 1 - basics
 ===================
 
-This tutorial demonstrates how to use ``doit`` to create simple computational pipeline.
+This tutorial demonstrates how to use ``doit`` to create a simple computational pipeline.
 
-The goal is to create a graph of python module imports in a package.
+The goal is to create a graph of python module's imports in a package.
 `graphviz <http://graphviz.org/>`_'s **dot** tool will be used to generate the graph.
 
-As example the `requests <https://github.com/requests/requests>`_ package will be used. The result is the image below:
+As an example, the `requests <https://github.com/requests/requests>`_ package will be used. The result is the image below:
 
 .. image:: _static/requests.png
 
@@ -21,12 +21,12 @@ In the image an arrow represents an import from one module to another. For examp
    from .structures import LookupDict
 
 
-Drawing an "import" dependency graph can be very useful to understand how the code in a package is organized.
+Drawing an "import" dependency graph can be very useful in understanding how the code in a package is organized.
 It may also help you identify potential problems like circular imports.
 
 There are three main steps to generate this graph:
 
-1) read each python module and list it's imports
+1) read each python module and list its imports
 2) generate a `dot` (text format) file representing the graph
 3) generate an image (PNG) from the `dot` file
 
@@ -88,10 +88,10 @@ The output contains one imported module per line.
 doit task
 ---------
 
-Next step we are going to wrap the above script in a ``doit`` *task*.
+On the next step we are going to wrap the above script in a ``doit`` *task*.
 In ``doit`` tasks are defined in a plain python module, by default called ``dodo.py``.
 
-For example a trivial task to execute the script above and save its output into a file would be:
+For example, a trivial task to execute the script above and save its output into a file would be:
 
 
 .. code-block:: python3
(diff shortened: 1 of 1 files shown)


doc/tutorial_1.rst
```diff
@@ -498,24 +498,28 @@ Let's rewrite the ``imports`` task to use python action:
...
 
 
-Function ``get_imports`` is used as task's action.
-It returns a dictionary, this will be saved by ``doit`` in its internal database. The returned dictionary must contain only values that can be encoded in JSON.
+The function ``get_imports`` is used as the task's action.
+It returns a dictionary, which will be saved by ``doit`` in its internal database. The returned dictionary must contain only values that can be encoded as JSON.
 
-Note that the parameter ``module_path`` is passed into task definition of ``actions``.
-Instead of just specifying a callable it takes a tuple *(callable, args, kwargs)*.
+``get_imports`` takes the path's module as a parameter. The value of this parameter to be used upon task execution is specified in the ``actions`` of the task definition. Generally speaking, each element of the ``actions`` array is a tuple *(callable, args, kwargs)*.
```
schettino72: Can you still include the parameter name, like:

``get_imports`` takes the path's module as a parameter (``module_path``).

doc/tutorial_1.rst
```diff
@@ -596,10 +600,10 @@ Sub-tasks (items of task group) by default are not reported on ``list`` command.
    imports:requests.utils
 
 
-Note the task name is composed by the (base name) group task name
```
schettino72: Here I have not introduced it, but a `doit` task might have an attribute called `basename`.
So I would like to keep consistent with `doit` terminology.
I suggest we use your alteration but still mentions "basename", something like:
 
Note the task name is composed by the task's group name (also known as ``basename``).

schettino72:
Also interested to get feedback about the tutorial itself (as being an introduction to the tool).
So please let me know what you think about it.

facundofc:
> Thank you so much for this. Very nice to get throughout review.
> I agree that all your changes improve the text.
> I just have a few comments regarding terminology when related to `doit`.
> 
> Feel free to add your name to AUTHORS file. I think you deserve some credit :)

Hey man! Thanks a lot for the nice feedback, it's really encouraging. Just wanted you to know I'll work on the changes you proposed, just give me some time.

facundofc:
> Also interested to get feedback about the tutorial itself (as being an introduction to the tool).
> So please let me know what you think about it.

I think the tutorial is well intended, but is still missing some love. On occasions it sounds like a list of steps, too recipe-like. I would try to give it a more personal touch, like if you were talking to the reader (this is, of course, just a matter of personal taste).

I don't know if you considered it or not, but perhaps it would be nice to pour in some comparisons with ``make``, for the (potentially large) group of users that already know ``make`` and are learning ``doit``. It's always nice to have a baseline you can hold on to and then take note of the differences/similarities. Maybe this could be implemented as some "If you know ``make``" notes spread through the document.

A bit more motivation at the beginning would be appropriate too. Not regarding why you'd like to make an imports' dependency graph, but why would you do it using ``doit``. I think this would be very similar to the Use Cases section (to some extent), but phrased different and intended for the users that directly skip to the tutorial.

That's all I can think of at the moment. If you agree on some of these, I volunteer to start making those changes. Just let me know.

Cheers.

schettino72:
> On occasions it sounds like a list of steps, too recipe-like. 

Agree, my writing style is very boring :grimacing: 
Would be nice to improve it to something more engaging.
Sometime I try but always think it sounds too cheesy and then I go back to my boring style. hehehehe

> I don't know if you considered it or not, but perhaps it would be nice to pour in some comparisons with make, 

I have considered this. The problem is that it adds noise to people who do not know `make`.
And most people who "knows" `make` actually never wrote a `Makefile`, they just know they have to execute `make` when compiling some code. I got thousands of people who ask me why they would need `doit` if python does not need compilation... And for people who actually understand `make` IMO would be easy for then to draw some comparisons. For those, maybe what would help is to point where `doit` is actually **different** from `make`.

So in general I am bit sceptic that this would be an improvement to write much in comparison to make. I think 2 or 3 notes regarding its differences might be ok.

> I think this would be very similar to the Use Cases section (to some extent), but phrased different and intended for the users that directly skip to the tutorial.

I try to not repeat much content, specially introduction, I am afraid to bore people who actually read the stuff... I mean I get annoyed when some docs keeps claiming the same thing but doesnt go into the HOW.

Actually when I wrote the tutorial I pretty much plan it based on the list of use-cases.
Maybe we could be more explicit about how each step correlates to a use-case.

> That's all I can think of at the moment. If you agree on some of these, I volunteer to start making those changes. Just let me know.

Thanks so much. In general I agree with the changes but see a few concerns above.
I **guess** I am receptive to changes but hard to say 100% before it is done.

Note that the tutorial is kind of new compared to the main documentation, the use-case section was not written by me. Of course improvements in any area are welcome...

cheers

PR #290, opened 2019-04-19
Fix development setup instructions

xordspar0:
dev/bin/activate isn't an executable script, you need to source the file
into your local environment. Following the instructions as they are now
results in a "not executable by this user" error.


schettino72:
thanks.

PR #292, opened 2019-04-27
Fix pathlib bug in tutorial

swenzel:
Hey there,

Today I went through your tutorial and noticed a few things:
First I had to install the system package `graphviz-dev` for `pygraphviz` to be installed correctly. Not your fault, but I thought mentioning it will spare other people some time.
Then `import-dep` didn't run because there was a bug in pathlib usage. I've fixed that and opened a pull request there already.
Then I saw the same bug in the tutorial.

At first I thought I'd just open an issue, but then I thought it's so easy to fix that I might just as well *doit* myself. Less complaining, more doing :smile:

Anyway, here you go!
I've also fixed a few other things that I noticed along the way but not all of them.
The sample `report_deps.py` doesn't work and spellchecking also fails on a few files. I think `hunspell` should also be mentioned as a system dependency for working on the docs somewhere.
Those are rather unrelated and I don't really know what's the right way to fix them which is why I left them unsolved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,5 @@ doc/samples/*.in
 doc/samples/*.out
 doc/samples/file*
 
-
+dev
+.vscode
--- a/doc/tutorial/tuto_1_1.py
+++ b/doc/tutorial/tuto_1_1.py
@@ -18,7 +18,7 @@ def module_to_dot(dependencies, targets):
     for dep in dependencies:
         filepath = pathlib.Path(dep)
         source = filepath.stem
-        with open(filepath) as fh:
+        with filepath.open() as fh:
             for line in fh:
                 sink = line.strip()
                 if sink:
--- a/doc/tutorial_1.rst
+++ b/doc/tutorial_1.rst
@@ -41,6 +41,7 @@ required python packages
 
   $ pip install doit pygraphviz import_deps
 
+Note that on some linux systems it is necessary to install the system package `graphviz-dev` first.
 
 sample project
 --------------
--- a/doc/tutorial_1.rst
+++ b/doc/tutorial_1.rst
@@ -622,7 +622,7 @@ key is the sub-task name:
 .. literalinclude:: tutorial/tuto_1_3.py
    :language: python3
    :lines: 25-41
-   :emphasize-lines: 5,15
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -14,4 +14,5 @@ doc/samples/*.in
 doc/samples/*.out
 doc/samples/file*
 
-
+dev
```
schettino72: Which software uses this?

Usually it is better to set editor ignore files in your user's *gitignore* file. Like `~/.gitignore`

doc/tutorial/tuto_1_1.py
```diff
@@ -18,7 +18,7 @@ def module_to_dot(dependencies, targets):
     for dep in dependencies:
         filepath = pathlib.Path(dep)
         source = filepath.stem
-        with open(filepath) as fh:
```
schettino72: What is the error you were getting? `open` is a valid python built-in function...

.gitignore
swenzel: `dev` is from doit, in your README you suggest to do 'virtualenv dev'.
`.vscode` is from VisualStudio Code. But putting that one in a personal gitignore sounds like a good idea. Never thought about that. I'll remove that line later.

doc/tutorial/tuto_1_1.py
swenzel: Yes it is, however it only supports `pathlib.Path` objects since python 3.6. I was using python 3.5.2 where this raises a `TypeError`. From your AppVeyor setup I see that you plan to support python3.4 and 3.5 as well so you'd need this for compatibility.

I've used `open` with `Path` objects myself and always got that `TypeError`, at least that's what I thought.
Until this morning, when I found out that it actually _does_ work with >=3.6. That was very confusing for me since I was certain that I've tried with 3.6 and 3.7 already. Anyway, so its not a bug but a compatibility issue.

schettino72: ah. ok... it is bug.

.gitignore
schettino72: I actually dont do it like that... I use virtualenvwrapper (or more recently virtualfish).

schettino72:
thanks

Issue #286, opened 2019-03-24
Cannot use functools.partial as title


slaperche-scality:
I was surprised to see that you can pass normal functions and lambas to the `title` attribute,but partial functions (`functools.partial`) aren't accepted:

```python
#!/usr/bin/env python
# coding: utf-8

import functools

def show_cmd(extra, task):
    return '{}: executing... {}'.format(extra, task.name)

def task_lambda_ok():
    return {
        'actions':['echo lambda'],
        'title': lambda task: show_cmd('LAMBDA', task)
    }

def task_partial_ko():
    title = functools.partial(show_cmd, extra='PARTIAL')
    return {
        'actions':['echo partial'],
        'title': title,
    }
```

gives

```
% doit list
ERROR: Task 'partial_ko' attribute 'title' must be {function, None}
got:functools.partial(<function show_cmd at 0x7f26847466a8>, extra='PARTIAL')
<class 'functools.partial'>
```

Is that on purpose or is it an oversight?

Seems prevented by https://github.com/pydoit/doit/blob/master/doit/task.py#L156, so it probably could be easily fixed by using `collections.abc.Callable` instead of `types.FunctionType` (but maybe I'm missing something).

schettino72:
> Is that on purpose or is it an oversight?

You are right. Oversight again. Trivial fix as you pointed out..

PR #293, opened 2019-05-01
Solve issue #286 - task.title accept partials

facundofc:
Had to switch to using `isinstance` in `Task.check_attr` to be able to use `collections.abc.Callable`. I think this change makes sense in general and not only to solve #286.

The minimum dodo file provided in #286's description now works (with minor modifications: had to swap `task` and `extra` parameters to `show_cmd`):

```python
#!/usr/bin/env python
# coding: utf-8

import functools

def show_cmd(task, extra):
    return '{}: executing... {}'.format(extra, task.name)

def task_lambda_ok():
    return {
        'actions':['echo lambda'],
        'title': lambda task: show_cmd(task, 'LAMBDA')
    }

def task_partial_ko():
    title = functools.partial(show_cmd, extra='PARTIAL')
    return {
        'actions':['echo partial'],
        'title': title,
    }
```

### DIFF
--- a/doit/task.py
+++ b/doit/task.py
@@ -407,9 +407,8 @@ class Task(object):
         @param valid (list): of valid types/value accepted
         @raises InvalidTask if invalid input
         """
-        for t in valid[0]:
-            if isinstance(value, t):
-                return
+        if isinstance(value, valid[0]):
+            return
         if value in valid[1]:
             return
 
--- a/doit/task.py
+++ b/doit/task.py
@@ -407,7 +407,7 @@ class Task(object):
         @param valid (list): of valid types/value accepted
         @raises InvalidTask if invalid input
         """
-        if isinstance(value, valid[0]):
+        if isinstance(value, tuple(valid[0])):
             return
         if value in valid[1]:
             return
--- a/doit/task.py
+++ b/doit/task.py
@@ -1,11 +1,11 @@
 
 """Tasks are the main abstractions managed by doit"""
 
-import types
 import os
 import sys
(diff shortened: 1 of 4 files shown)


doit/task.py
```diff
@@ -407,8 +407,9 @@ def check_attr(task, attr, value, valid):
         @param valid (list): of valid types/value accepted
         @raises InvalidTask if invalid input
         """
-        if type(value) in valid[0]:
-            return
+        for t in valid[0]:
```
schettino72: for-loop not required. `isinstance()` second attribute can be a tuple of types.

facundofc: Didn't know! Changed it, thanks!

doit/task.py
```diff
@@ -407,7 +407,7 @@ def check_attr(task, attr, value, valid):
         @param valid (list): of valid types/value accepted
         @raises InvalidTask if invalid input
         """
-        if isinstance(value, valid[0]):
+        if isinstance(value, tuple(valid[0])):
```
schettino72: why need to "convert" to tuple. they are all tuples already, right?

schettino72: oh. i see. problem is in tests itself. not in the code here...

facundofc: You're right! I apologize, I did the change in a rush while at work. I think the tests should be changed to use tuples and revert the convertion to tuple here, don't you?

Do you prefer to keep cumulating commits or should I rebase and `push -f`?

schettino72:
thanks

Issue #288, opened 2019-04-08
clean not able to access doit data?


moltob:
Actions are able to access doit's database, e.g. the return values of other tasks via `getargs`. As such configuration affects the action and what is done and produced, I see frequent need to also provide this to clean. But it does not seem to be possible. Is it?

Thx!

schettino72:
`clean` wont be able to access `getargs` because `clean` does not follow normal execution task's dependencies... but i understand your problem. You just need to save the data in the task that will use it.

Can you provide a simple example. I guess this would be a nice addition to (planned) HOW-TO section in docs.

moltob:
```python
import doit.tools

DOIT_CONFIG = dict(
    verbosity=2,
    default_tasks=['create_object'],
)


def task_determine_identifier():
    """Determine an identifier to be used by other tasks."""

    def do_complex_stuff():
        # Do some very complex or lengthy operation, like talking to an external service.
        ...
        return dict(identifier='0815')

    return dict(
        actions=[do_complex_stuff],
    )


def task_create_object():
    """Create object with ID."""

    # We don't know the identifier here (during task creation).

    def action(ident):
        print('Created object with ID', ident)

    def clean():
        # How to get identifier here?
        print('Deleted object with ID ???')

    return dict(
        actions=[action],
        getargs=dict(ident=('determine_identifier', 'identifier')),
    )
```

The second tasks needs values from the first task to define details of the action and also details of the corresponding clean.

schettino72:
Your example is pretty much what I expected. :smile:  

Running your example I found a bug on `result_dep` (used implicitly by `getargs`): fixed https://github.com/pydoit/doit/commit/3873759424a5a9559d7b6c8660bd72e0091bb0fc

I thought it was possible but actually currently not possible to do your use-case without changes on doit.
 
The problem is that `dep_manager` needs to be somehow exposed to task's action execution.
 We need to define an API... For `uptodate` functions, it can receive a magic `values` parameter that `doit` will inject saved values. We could use same solution here or maybe something more generic where whole `dep_manager` is exposed. What do you think? (Not sure you are familiar enough with doit internals)

I got it working with a quick and dirty patch...

```diff
diff --git a/doit/cmd_clean.py b/doit/cmd_clean.py
index 00d8d93..82845eb 100644
--- a/doit/cmd_clean.py
+++ b/doit/cmd_clean.py
@@ -57,6 +57,7 @@ class Clean(DoitCmdBase):
         for task in tasks:
[...]

PR #294, opened 2019-05-03
Taskloader API Update

moltob:
One way to implement #288 is the ability to access the dependency manager and all values stored by it from task related user-code, including during task generation. This PR is part 1 of that implementation and splits up the existing `task, config = TaskLoader.load_tasks(...)` API into two steps: configuration loading and actual task loading. This enables the dependency manager to be created _before_ task generators are called. The follow up PR will then add the dependency manager access methods.

Some remarks, how this implementation differs from the last pseudo code in #288:

* `ModuleTaskLoader` does not distinguish between dictionary and module in `setup`, but already in `__init__`. This prevents storing the additional attribute `mod_dict` in the instance, which also would be semantically redundant (and possibly a source of inconsistency).

### DIFF
--- a/doc/extending.rst
+++ b/doc/extending.rst
@@ -31,10 +31,10 @@ Normally `doit` tasks are defined in a `dodo.py` file.
 This file is loaded, and the list of tasks is created from
 the dict containing task meta-data from the *task-creator* functions.
 
-Subclass TaskLoader to create a custom loader:
+Subclass ``TaskLoader2`` to create a custom loader:
 
-.. autoclass:: doit.cmd_base.TaskLoader
-   :members: load_tasks
+.. autoclass:: doit.cmd_base.TaskLoader2
+   :members:
 
 
 The main program is implemented in the `DoitMain`. It's constructor
--- a/doc/extending.rst
+++ b/doc/extending.rst
@@ -36,6 +36,8 @@ Subclass ``TaskLoader2`` to create a custom loader:
 .. autoclass:: doit.cmd_base.TaskLoader2
    :members:
 
+Before the introduction of ``TaskLoader2`` a now deprecated loader interface ``TaskLoader`` was
+used, which did not separate the setup, configuration loading and task loading phases explicit.
 
 The main program is implemented in the `DoitMain`. It's constructor
 takes an instance of the task loader to be used.
--- a/doc/samples/custom_loader.py
+++ b/doc/samples/custom_loader.py
@@ -3,21 +3,23 @@
 import sys
 
 from doit.task import dict_to_task
-from doit.cmd_base import TaskLoader
+from doit.cmd_base import TaskLoader2
 from doit.doit_cmd import DoitMain
 
 my_builtin_task = {
     'name': 'sample_task',
     'actions': ['echo hello from built in'],
     'doc': 'sample doc',
-    }
+}
 
-class MyLoader(TaskLoader):
-    @staticmethod
-    def load_tasks(cmd, opt_values, pos_args):
+
+class MyLoader(TaskLoader2):
+    def load_doit_config(self):
+        return {'verbosity': 2}
+
+    def load_tasks(self, cmd, pos_args):
         task_list = [dict_to_task(my_builtin_task)]
-        config = {'verbosity': 2}
-        return task_list, config
+        return task_list
 
 
 if __name__ == "__main__":
--- a/doc/samples/custom_loader.py
+++ b/doc/samples/custom_loader.py
@@ -14,6 +14,9 @@ my_builtin_task = {
 
 
 class MyLoader(TaskLoader2):
+    def setup(self, opt_values):
+        pass
+
     def load_doit_config(self):
         return {'verbosity': 2}
 
--- a/doit/cmd_base.py
+++ b/doit/cmd_base.py
@@ -326,10 +326,6 @@ class TaskLoader2(TaskLoaderBase):
     """
     API = 2
 
-    def __init__(self):
-        super().__init__()
-        self.namespace = None
-
     def setup(self, opt_values):
         """Delayed initialization.
 
@@ -345,7 +341,7 @@ class TaskLoader2(TaskLoaderBase):
 
         :return: (dict) Dictionary of doit configuration values.
         """
-        return loader.load_doit_config(self.namespace)
+        raise NotImplementedError()
 
     def load_tasks(self, cmd, pos_args):
         """Load tasks.
@@ -356,10 +352,27 @@ class TaskLoader2(TaskLoaderBase):
         :param pos_args: (list str) positional arguments from command line
         :return: (List[Task])
         """
+        raise NotImplementedError()
+
+
+class NamespaceTaskLoader(TaskLoader2):
+    """Implementation of a loader of tasks from an abstract namespace.
+
+    A namespace is simply a dictionary to objects like functions and objects. See the derived
+    classes for some concrete namespace types.
+    """
+    def __init__(self):
+        super().__init__()
+        self.namespace = None
+
+    def load_doit_config(self):
+        return loader.load_doit_config(self.namespace)
+
+    def load_tasks(self, cmd, pos_args):
         return loader.load_tasks(self.namespace, self.cmd_names, cmd.execute_tasks)
 
 
-class ModuleTaskLoader(TaskLoader2):
+class ModuleTaskLoader(NamespaceTaskLoader):
     """load tasks from a module/dictionary containing task generators
     Usage: `ModuleTaskLoader(my_module)` or `ModuleTaskLoader(globals())`
     """
@@ -371,7 +384,7 @@ class ModuleTaskLoader(TaskLoader2):
             self.namespace = mod_dict
 
 
-class DodoTaskLoader(TaskLoader2):
+class DodoTaskLoader(NamespaceTaskLoader):
     """default task-loader create tasks from a dodo.py file"""
     cmd_options = (opt_dodo, opt_cwd, opt_seek_file)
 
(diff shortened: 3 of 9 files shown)


moltob:
@schettino72 Tests have been written and docs are adapted as far as I was aware of. PR ready for review.

doc/samples/custom_loader.py
```diff
@@ -3,21 +3,23 @@
...
     'actions': ['echo hello from built in'],
     'doc': 'sample doc',
-    }
+}
 
-class MyLoader(TaskLoader):
-    @staticmethod
-    def load_tasks(cmd, opt_values, pos_args):
+
+class MyLoader(TaskLoader2):
```
schettino72: I think better inlcude ``setup()`` in the example for documentation purpose. Even if just ``pass``...

doit/cmd_base.py
```diff
@@ -310,30 +310,64 @@ def _load_from(cmd, namespace, cmd_list):
         return task_list, doit_config
 
 
-class ModuleTaskLoader(TaskLoader):
+class TaskLoader2(TaskLoader):
```
schettino72: Why ``TaskLoader2`` subclass from ``TaskLoader``? I think better not.

doit/cmd_base.py
```diff
@@ -310,30 +310,64 @@ def _load_from(cmd, namespace, cmd_list):
...
+
+    The default implementation assumes tasks are loaded from a namespace, mapping identifiers to
+    elements like functions (like task generators) or constants (like configuration values).
+
+    This API update separates the loading of the configuration and the loading of the actual tasks,
+    which enables additional elements to be available during task creation.
+    """
+    def __init__(self):
+        super().__init__()
+        self.namespace = None
```
schettino72: I know I am the one who suggested to remove ``NamespaceLoader``... but **now** I am thinking it makes sense to have it. Sorry.

I guess it would make it clear that this ``namespace`` attribute is not required by the API.

doit/cmd_base.py
```diff
@@ -310,30 +310,64 @@ def _load_from(cmd, namespace, cmd_list):
...
+    def setup(self, opt_values):
+        """Delayed initialization.
+
+        To be implemented if the data is needed by derived classes.
+
+        :param opt_values: (dict) with values for cmd_options
+        """
+
+    def load_doit_config(self):
+        """Load doit configuration."""
```
schettino72: Please add note about return value ``dict``

doit/cmd_base.py
```diff
@@ -310,30 +310,64 @@ def _load_from(cmd, namespace, cmd_list):
...
+        return loader.load_doit_config(self.namespace)
+
+    def load_tasks(self, cmd, pos_args):
+        """Load tasks.
+
+        :return: (List[Task])
```
schettino72: Not sure it is a standard but usually I put ``return`` description *after*  ``param``

doit/cmd_base.py
```diff
@@ -310,30 +310,64 @@ def _load_from(cmd, namespace, cmd_list):
...
+
+
+class ModuleTaskLoader(TaskLoader2):
     """load tasks from a module/dictionary containing task generators
     Usage: `ModuleTaskLoader(my_module)` or `ModuleTaskLoader(globals())`
     """
-    cmd_options = ()
-
     def __init__(self, mod_dict):
         super(ModuleTaskLoader, self).__init__()
```
schettino72: since we are touching this code, lets use modern ``super`` (without parameters)

doit/cmd_base.py
```diff
@@ -310,30 +310,64 @@ def _load_from(cmd, namespace, cmd_list):
...
-            params['cwdPath'],
-            params['seek_file'])
-        return self._load_from(cmd, dodo_module, self.cmd_names)
+    def setup(self, opt_values):
+        # lazily load namespace from dodo file per config parameters:
+        self.namespace = loader.get_module(
+            opt_values['dodoFile'],
+            opt_values['cwdPath'],
+            opt_values['seek_file'],
+        ).__dict__
```
schettino72: Original code is using ``inspect.getmembers``. Any subtle difference? I guess better not rely on ``__dict__``.

doit/cmd_base.py
```diff
@@ -446,8 +480,17 @@ def execute(self, params, args):
         :param params: instance of cmdparse.DefaultUpdate
         :param args: list of string arguments (containing task names)
         """
-        self.task_list, dodo_config = self.loader.load_tasks(
-            self, params, args)
+
+        # distinguish legacy and new-style task loader API when loading tasks:
+        if isinstance(self.loader, TaskLoader2):
```
schettino72:
```
legacy_loader = not  isinstance(self.loader, TaskLoader2):
if legacy_loader:
   ....
else:
  ....
```

One line shorter :grin: , and more readable IMO.

schettino72: I would prefer we add an ``API=2`` check, so users are not forced to be subclass ``TaskLoader2`` as long as they implement the interface...

doit/cmd_completion.py
```diff
@@ -98,8 +97,12 @@ def _generate_bash(self, opt_values, pos_args):
 
         # if hardcode tasks
         if opt_values['hardcode_tasks']:
-            self.task_list, _ = self.loader.load_tasks(
-                self, opt_values, pos_args)
+            if isinstance(self.loader, TaskLoader2):
+                self.loader.setup(opt_values)
+                self.task_list = self.loader.load_tasks(cmd=self, pos_args=pos_args)
```
schettino72: Uhhmm. I have not thought about the possibility of calling ``load_tasks`` without (or before) calling `` load_doit_config``.

Not sure about it. I think this should be forbidden somehow. ``load_doit_config`` might perform some configuration changes in the loader itself... What do you think?

tests/test_cmd_base.py
```diff
@@ -121,15 +121,26 @@ def test_failCall(self):
...
         loader = ModuleTaskLoader(members)
-        task_list, config = loader.load_tasks(cmd, {}, [])
+        loader.setup({})
+        config = loader.load_doit_config()
+        task_list = loader.load_tasks(cmd, [])
+        assert ['xxx1'] == [t.name for t in task_list]
+        assert {'verbose': 2} == config
+
+    def test_load_tasks_from_module(self):
+        import tests.module_with_tasks as module
```
schettino72: Please put ``import`` on top of file.

tests/test_cmd_base.py
```diff
@@ -193,6 +206,32 @@ def test_execute(self, depfile_name):
...
+        class LegacyLoader(TaskLoader):
+            def load_tasks(self, cmd, opt_values, pos_args):
+                return super()._load_from(cmd, members, [])
+
+        mycmd = self.MyCmd(task_loader=LegacyLoader())
+        assert 'min' == mycmd.parse_execute([
+            '--db-file', depfile_name,
+            '--mine', 'min',
+        ])
+
```
schettino72: **I** usually use two blank line between methods if the method as blank lines to separate blocks. Otherwise I use just one line. And if methods are separated by 2 blank lines, I use 3 between classes.

Not PEP8.  But I guess this makes code more readable. Although not always strict about it.

tests/test_cmd_strace.py
```diff
@@ -14,12 +14,21 @@
     "os.system('strace -V') != 0 or sys.platform in ['win32', 'cygwin']")
 class TestCmdStrace(object):
 
+    @staticmethod
+    def loader_for_task(task):
```
schettino72: I guess it would make more sense to just subclass ``TaskLoader2``instead of mocking everything...

doc/extending.rst
```diff
@@ -31,10 +31,10 @@ Normally `doit` tasks are defined in a `dodo.py` file.
 This file is loaded, and the list of tasks is created from
 the dict containing task meta-data from the *task-creator* functions.
 
-Subclass TaskLoader to create a custom loader:
```
schettino72: Since we are not even adding a deprecation warning (yet).
Can we just mention there is a "deprecated" older version?

doit/cmd_base.py
moltob: I usually don't touch existing code just for personal taste. PyCharm shows little inconsistencies to my code style settings everywhere, but as this might be personal, I don't try to convince. But sure, I'll adapt it for you.

doit/cmd_completion.py
moltob: I can add the config load call here as well. I would not go so far as to double-checking in the loader classes, I did not see anything like this done elsewhere.

tests/test_cmd_base.py
moltob: Are you sure? This is a module needed for this particular test function only. I strongly suggest to keep it local. It is not uncommon at all to have function local imports in test code.

Please let me know.

doit/cmd_base.py
schettino72: Yes, usually dont modify existing code. but this class is now a new class...

Which style inconsistencies? As far as I know I only deviate from PEP8  regarding blank lines.

tests/test_cmd_base.py
schettino72: ok, keep it local

doit/cmd_completion.py
schettino72: yes. double-checking might be overkill, but should be noted somewhere...

doit/cmd_base.py
moltob: Don't worry, as I said, this is always personal taste and I of course accept the library's style.

If you are really interested I suggest to run a style checker (I use one built into PyCharm) and a linter like pylint.

doit/cmd_completion.py
moltob: Ok, I'll add to docs as well.

moltob:
@schettino72 Review comments incorporated, please pick up the review when you get a chance.

schettino72:
Thanks. squashed and merged.

I removed BaseTaskLoader, and duplicated the code.... I chose to do this way because I expect the duplicated code to be removed in the future.

Issue #285, opened 2019-03-24
Unexpected targets' cleanup order


slaperche-scality:
I wanted a task that creates a directory hierarchy, so I was careful to sort my list of targets from the root to the leaves and I was expecting that the cleanup would be done in the reverse order (thus, leaves would be `rmdir`ed before the root but it seems the cleanup iterate over the targets from the start.

Given this `dodo.py`

```python
#!/usr/bin/env python
# coding: utf-8

import os

def task_test():
    """test"""
    def mkdir(targets):
        for directory in targets:
            os.mkdir(directory)

    return {
        'actions': [mkdir],
        'targets': ['foo', 'foo/bar'],
        'clean': True,
    }
```

I got

```
% doit clean test
test - cannot remove (it is not empty) 'foo'
test - removing dir 'foo/bar'
```

Now, it's easy to workaround (my action can iterate over `reversed(targets)` or I can write a custom cleanup action) but it's not ideal.

Is it a bug or was I having unjustified expectations here?

schettino72:
> Is it a bug or was I having unjustified expectations here?

You are right. Makes sense clean-up order is reversed.
It is already reverses tasks, not reversing the list of targets was an oversight.
Should be a trivial fix.

facundofc:
Hi there! I've been taking a look at the code that removes the targets ([here](https://github.com/pydoit/doit/blob/master/doit/task.py#L574)) and as you can see it's not even removing the targets in order. It actually splits the list into files and directories and removes them in that order. I track down this to the very old commit 97855edf.

I guess the rationale was to first remove the files so as to maybe then leave the directories empty and be able to delete them too, so that if your task has `'targets': ['a_path/a_file', 'a_path']` then the cleaning works.

I can think of two solutions:

1. We reverse both lists, `files` and `dirs`, but keep deleting first files and then directories.
2. We reverse the targets' list and iterate over it, testing whether current element `isdir` or `isfile` and deleting as appropriate.

Both solutions have the potential to break existing tasks, but I think option 1. should be preferred since it seems less likely to do so (if some figured out this splitting mechanism, then fixing such a task would just require to reverse the targets' list; while with solution two it might be necessary to put some thought into the targets' reordering to make the cleaning work again).

What's your opinion?

slaperche-scality:
Solution 2 has the advantage of being predictable (there is no "magic" that separate files and directories) and consistent (is this splitting also done for creation? Like, creating directories first and then files? If not, it may be confusing).

[...]

PR #295, opened 2019-05-05
Fix issue #285 - Unexpected targets' cleanup order

facundofc:


CHANGES
```diff
@@ -21,6 +21,7 @@ Changes
  - Add configuration `DOIT_CONFIG` `action_string_formatting` to control command action formatter.
  - Fix `result_dep`, use result **after** its execution
  - Fix #286: Cannot use functools.partial as title
+ - Fix #285: Unexpected targets' cleanup order
```
schettino72: CHANGES entries should read like "what was fixed", not like "issue description".
Same could be said about git commit

- Fix #286: Cannot use functools.partial as title
+ Fix #286: Support functools.partial on task's dict `metadata task.title`
- Fix #285: Unexpected targets' cleanup order
+ Fix #285: clean command, remove files before attempting to remove folders

tests/test_task.py
```diff
@@ -433,6 +437,19 @@ def test_keep_non_empty_dirs(self, tmpdir):
             assert expected == os.path.exists(filename)
         assert os.path.exists(tmpdir['dir'])
 
+    def test_clean_in_reverse_lexical_order(self, tmpdir):
```
schettino72: The test name is misleading because it actually does not ensure reverse lexical order.

facundofc:
> thanks.
> 
> sorry if review sounds too nitpick.
> I can do these changes later myself when merging.

Hey! I don't find it nitpick, I think it's understandable to wish to keep some consistency.

CHANGES
facundofc: Understood. I had the impression that the other changes quoted the issue's title, but I didn't actually go and check.

I changed the phrasing of #285 solution, let me know what you think.

tests/test_task.py
facundofc: I agree. I'm having a bit of a tough time finding a proper name though. An option I considered is to actually do ensure that the order is that one, by patching both `os.rmdir` and `os.remove`, what would you think about this?

CHANGES
```diff
@@ -20,7 +20,8 @@ Changes
  - Make it possible to use a custom encoder when using config_changed with a dict.
  - Add configuration `DOIT_CONFIG` `action_string_formatting` to control command action formatter.
  - Fix `result_dep`, use result **after** its execution
- - Fix #286: Cannot use functools.partial as title
+ - Fix #286: Support `functools.partial` on tasks' dict metadata `task.title`
+ - Fix #285: `clean` command, remove targets in reverse DFS order
```
schettino72: What is DFS? "Distributed File System"?

schettino72:
merged. thanks

CHANGES
facundofc: It's Depth First Search in this case. I didn't mean to be pedantic.

Thanks for merging!

PR #296, opened 2019-05-11
fix test warnings

moltob:
Unit tests raise these warnings:

```
============================== warnings summary ===============================
tests\test_cmd_completion.py:90
  C:\projects\doit\tests\test_cmd_completion.py:90: DeprecationWarning: invalid escape sequence \]
    assert ('"(-s|--lll)"{-s,--lll}"[my desc [b\]a]" \\' ==
tests\test_cmd_completion.py:97
  C:\projects\doit\tests\test_cmd_completion.py:97: DeprecationWarning: invalid escape sequence \]
    assert ('''"(-s|--lll)"{-s,--lll}"[my \\"des'c [b\]a]" \\''' ==
-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

This PR fixes the two invalid escape sequences by replacing `\]` with `\\]`.

### DIFF
--- a/tests/test_cmd_completion.py
+++ b/tests/test_cmd_completion.py
@@ -87,14 +87,14 @@ class TestCmdCompletionZsh(object):
 
         opt4 = CmdOption({'name':'o4', 'default':'', 'help':'my desc [b]a',
                           'short':'s', 'long':'lll'})
-        assert ('"(-s|--lll)"{-s,--lll}"[my desc [b\]a]" \\' ==
+        assert ('"(-s|--lll)"{-s,--lll}"[my desc [b\\]a]" \\' ==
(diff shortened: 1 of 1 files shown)


moltob:
@schettino72 Verified that warnings are gone on Linux and Windows, PR ready for review.

schettino72:
thanks. 

How do you see these warnings when running tests?

moltob:
Both CIs showed them, also see https://docs.pytest.org/en/latest/warnings.html.

schettino72:
ah ok. i see it. on my local machine it shows only once after i modify that line... tried to use -W without success, they always hide it after first run...

I guess i will change it to error, I usually dont read CI tracebacks unless something fails.

PR #297, opened 2019-05-12
Public dependency manager

moltob:
In order to implement https://github.com/pydoit/doit/issues/288, this PR stores the dependency manager in a global registry class before task are loaded or executed.

This allows to get it from any code inside a task generator, e.g. from clean to read the previous execution result, see the contained sample.

The original discussion suggested to implement the registry as plain Python `dict`. I'd like to prevent string-based access and therefore implemented a very simply static class in new module `doit.globals`.

### DIFF
--- a/doc/globals.rst
+++ b/doc/globals.rst
@@ -1,0 +1,31 @@
+==========
+Singletons
+==========
+
+During the lifecycle of a command invocation, some properties of `doit` are stored in singletons,
+provided by the ``doit.globals.Globals`` class.
+
+.. autoclass:: doit.globals.Globals
+   :members:
+
+dep_manager
+-----------
+
+The doit dependency manager holds the persistent state of `doit`. This includes relations of tasks
+among each other or results, returned from their latest runs. It bascially consists of all the
+information stored in `doit`'s database file.
+
+The ``dep_manager`` attribute is initialized right before tasks are loaded, which means it allows to
+be accessed during *all* task evaluation phases, in particular during:
+
+* Task generation, i.e. from the body of any ``task_*`` function. Task execution, i.e. from the code
+* executed by one of the task's actions. Task cleanup, i.e. from the the code executed by one of the
+* tasks's clean activities.
+
+An example of this is a task, where at generation time the *target* of the task action is not yet
+known, because it is determined during execution. Then it would be possible to store that target in
+the dependency manager by returning it from the action. A `clean` action is then able to query
+`dep_manager`for that result and perform the cleanup action:
+
+
+.. literalinclude:: samples/global_dep_manager.py
--- a/doc/globals.rst
+++ b/doc/globals.rst
@@ -18,9 +18,9 @@ information stored in `doit`'s database file.
 The ``dep_manager`` attribute is initialized right before tasks are loaded, which means it allows to
 be accessed during *all* task evaluation phases, in particular during:
 
-* Task generation, i.e. from the body of any ``task_*`` function. Task execution, i.e. from the code
-* executed by one of the task's actions. Task cleanup, i.e. from the the code executed by one of the
-* tasks's clean activities.
+* Task generation, i.e. from the body of any ``task_*`` function.
+* Task execution, i.e. from the code executed by one of the task's actions.
+* Task cleanup, i.e. from the the code executed by one of the tasks's clean activities.
 
 An example of this is a task, where at generation time the *target* of the task action is not yet
 known, because it is determined during execution. Then it would be possible to store that target in
--- a/doc/globals.rst
+++ b/doc/globals.rst
@@ -1,9 +1,9 @@
-==========
-Singletons
-==========
+=======
+Globals
+=======
 
-During the lifecycle of a command invocation, some properties of `doit` are stored in singletons,
-provided by the ``doit.globals.Globals`` class.
+During the lifecycle of a command invocation, some properties of `doit` are stored in gloabl
+singletons, provided by the ``doit.globals.Globals`` class.
 
 .. autoclass:: doit.globals.Globals
    :members:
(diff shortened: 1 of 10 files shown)


moltob:
@schettino72 Where would I link a sample snippet into the docs, what would be a good spot?

schettino72:
I will rebase this before reviewing it...

schettino72:
Actually quite complicated to rebase this to master... I gave up.
I think it is easier no manually apply your changes.  sorry for the trouble. I could do that myself later, not today.

Your ``Globals`` as Registry implementation is ok. I think the ``get_dep_manager`` turned out be useless. We can just ask user to import ``Globals`` and remove ``get_dep_manager``. What do you think?

Regarding the documentation, your example i would like to put in a new HOW-TO section. dont worry about that now, we can do it later. 

But `Globals` would need to be documented now, not sure where... here I would you use a simpler example (with just `actions` without `clean`).

moltob:
Don't worry, I'll fix the branch and will incorporate your comments.

moltob:
@schettino72 The diff is clean now, you can add your comments for the files, but I can easily just remove the access function as well. I'll do that tomorrow.

moltob:
I just rebased the code cleanly onto master locally, will push it tonight.

Regarding the sample I'd like to keep the clean. IMHO this is one of the main use cases for that feature. A slightly simplified sample will be part of the pushed branch.

moltob:
@schettino72 Please pick up review, rebase onto master is complete. Please see my above comments on having `clean` in the sample.

doit/globals.py
```diff
@@ -0,0 +1,9 @@
+"""Simple registry of singletons."""
+
+
+class Globals:
+    """Accessors to doit singletons.
+
+    :cvar dep_manager: The doit dependency manager, holding all persistent task data.
```
schettino72: Could you add the type here

tests/test_cmd_base.py
```diff
@@ -207,6 +208,23 @@ def test_execute(self, depfile_name):
             '--mine', 'min'])
 
 
+    @mock.patch('doit.cmd_base.Globals')
```
schettino72: Interesting, usually I use `pytest.monkeypatch`... didnt know about this.

tests/test_cmd_base.py
```diff
@@ -207,6 +208,23 @@ def test_execute(self, depfile_name):
...
+    def test_execute_provides_dep_manager(self, mock_globals, depfile_name):
+        mock_globals.dep_manager = None
+        members = {'task_xxx1': lambda: {'actions': []}}
+
+        class MockTaskLoader(ModuleTaskLoader):
+            def load_tasks(self, cmd, pos_args):
+                assert mock_globals.dep_manager, 'dep_manager not set before tasks are loaded'
```
schettino72: I think add an ``assert`` explanation is not a good practice with py.test.
py.test will rewrite the assert and give a better error message when there is no message.
I usually write a comment above the assert when it needs some kind of explanation.

moltob: I was not aware of [this](https://docs.pytest.org/en/latest/assert.html#assert-details). Thanks for the hint!

moltob:
@schettino72 Docs added.

doc/globals.rst
```diff
@@ -0,0 +1,31 @@
+==========
+Singletons
```
schettino72: the file is  called 'globals' the the title is 'singletons'... we should use just one name...

you write the docs as if we would have many globals, not sure that would be the case. maybe we could expose the Reporter or some other internal stuff. But those are already documented elsewhere...

what about have this section of docs being specific to 'db backend' or something like that.

Another thing i just realized is that we are calling it 'dep_manager' as it is original function on doit itself.Wondering if we should use a more generic name. In practice users could use this to store some data. like a a task that stores and use a data itself in future runs. for  this use-case the name dep_manager is completely innapropriate.

what do you think?

doc/globals.rst
```diff
@@ -0,0 +1,31 @@
...
+* Task generation, i.e. from the body of any ``task_*`` function.
+* Task execution, i.e. from the code executed by one of the task's actions.
+* Task cleanup, i.e. from the the code executed by one of the tasks's clean activities.
+
+An example of this is a task, where at generation time the *target* of the task action is not yet
+known, because it is determined during execution. Then it would be possible to store that target in
+the dependency manager by returning it from the action. A `clean` action is then able to query
+`dep_manager`for that result and perform the cleanup action:
+
+
```
schettino72: description looks good. but we also need to document the exposed API

doc/globals.rst
```diff
@@ -0,0 +1,31 @@
...
+.. autoclass:: doit.globals.Globals
+   :members:
+
+dep_manager
+-----------
+
+The doit dependency manager holds the persistent state of `doit`. This includes relations of tasks
+among each other or results, returned from their latest runs. It bascially consists of all the
+information stored in `doit`'s database file.
```
schettino72: maybe  add a note about doit different backends...

doc/globals.rst
moltob: I'll adapt the section title to match file and module name.

> you write the docs as if we would have many globals

Correct, this way, it will be easy to extend it when needed. I certainly would not add anything here just to have more than one attribute.

Regarding the names: I surely stumbled upon the names initially as well. There is a class `Dependency` that is called `dep_manager` wherever it is used and in fact it is actually the doit database...

But I got over it 😃  If you want to rename it and clean this up, I'd ask you to do this in a separate PR, if you don't mind. I maybe would have called things differently here and there, but that is really a personal thing...

moltob:
Added docs about dep_manager API, some of the backend API and the different backends.

schettino72:
Merged, thanks. Nice to have a regular contributor :+1:

Notes:

- There is a spell checker task for docs `doit spell` (requires installation of hunspell).
- on docs I like breaking lines on period, etc. Makes it easier to review changes and not having very long lines (still max 80 column per line)
- your sample was broken. there is also a task to check them `doit samples_check` (not sure they work on Windows).
- after all I guess the name `dep_manager` is ok.
- I also exposed Globals as `doit.Globals`, `doit.globals.Globals` was looking like Java :stuck_out_tongue:

Issue #98, opened 2015-08-18
add support for non-serializable `getargs` 


schettino72:
See initial discussion on: https://groups.google.com/d/topic/python-doit/c1ZK4GNJkSw/discussion

One important consideration is that task that creates non-serializable values 
can never be considered up-to-date,
otherwise a task that uses this value would fail.

There are 2 ways to implement this:

1) The action would wrap the returned value in a custom doit Object that indicates it should not be serialized. It will be up to the user to make sure the task is never up-to-date.

2) add a new attribute to a task to indicate its result/values are not serializable.

schettino72:
@FrankStain I suggested option 1 in the email... but thinking over I guess option 2 would be better. What do you think? What would work best for your use-case?

FrankStain:
@schettino72 First variant looks more flexible as for doit customers as for it's development.
Looks like a kind of envelope where object is stored by one task to be taken by another.

And about 'never-up-to-date' i can say that currently i need to use 'uptodate': [False] for all such tasks to schedule it's execution every pipeline launch. So it's good idea consider such tasks to be never up-to-date.

schettino72:
This is done but docs are too superficial:

https://pydoit.org/cmd_run.html#codec-cls

rbdixon:
OK, I'll fix up the docs in January. Thanks for taking a look.

PR #301, opened 2019-05-15
Implement configurable JSON encoder/decoder to allow non-serializable python-action return values

rbdixon:
Per #98.

- [x] Add encoder/decoder parameters to JSON backend
- [x] Add encoder/decoder parameters to gdbm backend
- [x] Add encoder/decoder parameters to sqlite backend
- [x] Add encoder/decoder parameter to `Dependency` manager
- [x] Add tests for backends
- [x] Add `get_encoder_decoder()` function to `DoitCmdBase` to use config parameters to specify JSON encoder/decoder
- [x] Pass encoder/decoder value to `Dependency` manager instance in `cmd_base.py`
- [x] Document new configuration parameter

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -97,16 +97,18 @@ def dep_manager_fixture(request, dep_class, encoder_cls=json.JSONEncoder, decode
     dep_file.whichdb = whichdb(dep_file.name) if dep_class is DbmDB else 'XXX'
     dep_file.name_ext = db_ext.get(dep_file.whichdb, [''])
 
+    return dep_file
+
+
+@pytest.fixture
+def dep_manager(request):
+    dep_file = dep_manager_fixture(request, DbmDB)
+
     yield dep_file
 
     if not dep_file._closed:
         dep_file.close()
     remove_db(dep_file.name)
-
-
-@pytest.fixture
-def dep_manager(request):
-    return dep_manager_fixture(request, DbmDB)
 
 
 @pytest.fixture
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,9 +14,9 @@ def get_abspath(relativePath):
     """ return abs file path relative to this file"""
     return os.path.join(os.path.dirname(__file__), relativePath)
 
+
 # fixture to create a sample file to be used as file_dep
 def dependency_factory(relative_path):
-
     @pytest.fixture
     def dependency(request):
         path = get_abspath(relative_path)
@@ -29,6 +29,7 @@ def dependency_factory(relative_path):
         def remove_dependency():
             if os.path.exists(path):
                 os.remove(path)
+
         request.addfinalizer(remove_dependency)
 
         return path
@@ -45,9 +46,11 @@ def target1(request):
     path = get_abspath("data/target1")
     if os.path.exists(path):  # pragma: no cover
         os.remove(path)
+
     def remove_path():
         if os.path.exists(path):
             os.remove(path)
+
     request.addfinalizer(remove_path)
     return path
 
@@ -68,6 +71,7 @@ def remove_db(filename):
         if os.path.exists(filename + ext):
             os.remove(filename + ext)
 
+
 # dbm backends use different file extensions
 db_ext = {
     'dbhash': [''],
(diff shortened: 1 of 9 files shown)


tests/conftest.py
```diff
@@ -76,27 +76,31 @@ def remove_db(filename):
...
+    name = request._pyfuncitem.name
+    name = py.std.re.sub("[\W]", "_", name)
+    return request.config._tmpdirhandler.mktemp(name, numbered=True)
+
+@pytest.fixture
+def dep_manager(request, dep_tmpdir):
+    # TODO: This doesn't work as expected
+    # Migrate to test_dependency.py fixture
     if hasattr(request, 'param'):
+        raise Exception('Did not expect this code to be used... investigate')
```
rbdixon: @schettino72 I _think_ this fixture was attempting to inspect the `request.param` attribute which is in use on the fixture parametrization on `test/test_dependency.py#69` and then setup the requested backend. pytest does not currently work like that. Maybe it used to?

To verify I put that exception in just to see if anything hit it. Nothing ever does.

I duplicated this fixture in `tests/test_dependency.py` and moved the backend setup code there.

I think that in prior versions only the default backend was being tested in `tests/test_dependency.py`.

rbdixon:
This is usable and has working tests. I'm going to try it out here for a while and see how it goes.

tests/conftest.py
schettino72: good catch. thanks. fixed on master.

rbdixon:
Rebased. I'll squash this whole series of patches into one commit once it is ready.

schettino72:
thanks,  usually i squash myself when merging.

if you squash you mess up with inline comments here. and if requires another review you would need to sqaush again..  not productive...

rbdixon:
I think this is ready to go. I’ve been using it locally for a few days. 

I’m offline for a week so no rush merging it. If you would like me to do more on this let me know. Thank you!

schettino72:
It is not good practice to reformat the code... makes code-review harder. But I am guilty of having some really bad formatting there (apart from my decision to ignore PEP8 regarding vertical space), so I will fix the mess myself.

I thought it was interesting that you wanted to use pickle...  putting pickle inside JSON is ugly as hell, but why not support pickle as first class? I want to explore this and might spin this PR in a different direction.

Anyway I will handle it myself, enjoy your time off :smile:

rbdixon:
Oh... I’m sorry. I use a formatter on all my work projects and sometimes it creeps into non work projects. I can back that all out if you would prefer. 

If you ever want to consider a formatter black is very nice.

rbdixon:
I’m back from vacation. Anything I can help with to get this merged?

schettino72:
can you please rebase to master and remove the formatting changes?

on master i fixed some formatting problems.... i am not a fan.of 'black' for formatting as i use blank lines different from pep8 recommendation.

rbdixon:
OK, I think I've got all of the changes black made to the test directory reverted.

All tests pass _except_ these tests which also fail on master:

- `tests/test_cmd_auto.py:TestAuto.test_run_wait`
- `tests/test_filewatch.py:TestFileWatcher.testLoop`: This test seems to get stuck waiting.

I hope this is helpful. thank you!

rbdixon:
How did the `black` reversions look?

schettino72:
Sorry I had a review of this that was not published. I have merged the functionality, my changes:

- need to review docs ( I ran out of time, but merged anyway)
- I combined encoder/decoder in a single "codec" parameter (no need to be JSON based!)
- I removed support to pass a class as a string. I have already written code to support this for plugins. So this should be a plugin...
- Also possible to pass a class directly on DOIT_CONFIG, example below. 
- I removed JSONNullEncoder
- I dont think test_dependency really needs to test codecs...
- fixed strace tests that were broken


```
# encoder for db backend
class JSONNullEncoder(json.JSONEncoder):
    'Encodes non-serializable values as null'

    def default(self, obj):
        if type(obj) not in [str, int, float, bool, list, dict, None]:
            return None
        else:
            return obj

class MyCodec(JSONCodec):
    def __init__(self):
        self.encoder = JSONNullEncoder()
        self.decoder = json.JSONDecoder()

DOIT_CONFIG = {
    'codec_cls': MyCodec,
}
```

schettino72:
leaving this open until I merge docs.

rbdixon:
I gave this a try and it works as expected. Thank you.

The only problem I see is a minor typo in `CHANGES`: `BACKWARS INCOMPATIBLE`

This is really helpful for my application of doit.

PR #304, opened 2019-05-21
[minor] Docs: call attn to custom doc (besides docstring)

hangtwenty:
* Seeing that I could customize the `doc` attribute of a task, I realized I actually like subtasks. (Before I had found them confusing as it seemed like I had to choose between all-tasks docstring or nothing.)
* Might make sense to call this out to readers early on.

----

I figure this is not the only doc file to update, if the change is to be accepted... Let me know?

### DIFF
--- a/doc/samples/task_reusable.py
+++ b/doc/samples/task_reusable.py
@@ -1,9 +1,11 @@
 
 def gen_many_tasks():
+    """ first line of docstring becomes doc kwarg """
     yield {'basename': 't1',
            'actions': ['echo t1']}
     yield {'basename': 't2',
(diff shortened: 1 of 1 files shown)


schettino72:
Yes, I guess you are right? I think tje.problem is that docs take so long to introduce the 'list' command and 'doc' only makes sense for 'list'. maybe we should introduce 'list' early on somewhere on 'task.rst'.

schettino72:
Thanks for feedback. I thought it was better to use a separate example:
https://github.com/pydoit/doit/commit/c09361844f4c2f3003b66ee523d789eb0b3b70fc

PR #306, opened 2019-06-05
doc: add MetalK8s success story

slaperche-scality:
Here is the success story about how we replaced the existing `make`-based build system by `doit` on [MetalK8s](https://github.com/scality/metalk8s/).

Thanks for the great tool!

### DIFF
--- a/doc/stories.rst
+++ b/doc/stories.rst
@@ -183,6 +183,47 @@ Each sound should be loaded from PCM, converted into MP3 or OGG and linked with 
 `doit` is well designed tool for such purposes, i think.
 
 
+`MetalK8s @ Scality <https://www.scality.com/>`_
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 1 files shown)


schettino72:
Great, thanks :smile: 

> `Invoke`` was pretty good at executing commands

Is it better then ``doit`` in this aspect? Why? How doit could improve...

I am thinking about adding the logo of project/companies in doit ``homepage``, do you think it would ok to add it? Need any special permission?

moltob:
@schettino72 Regarding the BMW success story, showing the logo should be fine. I suggest to make the bitmap a link to the story for which I was given permission to release.

slaperche-scality:
@schettino72 

> Is it better then doit in this aspect? Why? How doit could improve...

Not really, at least for our use `doit` is as convenient as `Invoke` :slightly_smiling_face: 

> I am thinking about adding the logo of project/companies in doit homepage, do you think it would ok to add it? Need any special permission?

I asked around, and no problem on our side: you can add the MetalK8s logo on `doit` homepage.

Issue #311, opened 2019-06-29
command line arguments for task creators


schettino72:
`doit` already support Task parameters, but they are not convenient to be used with task-groups.

Example of proposed feature:

```
from doit import task_param

@task_param({'name':'param1', 'short':'p', 'default':'default value'})
def task_py_params(param1):
    for name in ['foo', 'bar']:
        yield {
            'name': name,
            'actions':['echo {}'.format(param1)],
        }
```

`doit` would be responsible for injecting parameter values when executing task-creator functions.

smarie:
I would love to see this working. As of now there does not seem to be a workaround, no ?

smarie:
Alternative : we could benefit from the fact that generators can `return` an argument (that is captured in the `StopIteration` exception):

```
from doit import task_param

def task_py_params(param1):
    for name in ['foo', 'bar']:
        yield {
            'name': name,
            'actions':['echo {}'.format(param1)],
        }
    return {'params': [{'name':'param1', 'short':'p', 'default':'default value'}]}
```

The doc would need to explain that in that case only some fields are allowed (I can only think of `'params'` for now but I'm sure there will ultimately be other things)

That would be closer to your initial syntaxt choices. Of course creating decorators for all aspects of task metadata information providing are also cool, but it is maybe another feature (not only for task generators). And it resembles very much what has been done in the [`letsdoit`](https://bitbucket.org/takluyver/letsdoit/src/default/) plugin

rbdixon:
@smarie See PR #323... I _think_ this is what was intended by the issue. Does this do what you were thinking, too?

smarie:
There seem to be several features behind this ticket:

 - first, the fact that a "group task" exists as a task in `doit`, even if the user is not aware of it today. We could offer the user with the capability to **edit that group task**. This is the direction proposed by PR #321, and it could go much further (adding actions executed at the end of the group for example).

 - second, the fact that subtasks in a group do not today have the possibility to **inherit from common group behaviour**, in particular parameter definition. This is also the direction proposed by PR #321: when the user defines a parameter on the group task, it is copied on all subtasks.

[...]

Issue #283, opened 2019-02-07
Set task options from doit.cfg


rbdixon:
Per [thread](https://groups.google.com/forum/#!topic/python-doit/4PnYOeNO5SI): I'd like to set task parameters from doit.cfg as well. It seems possible since the `doit help <task>` section lists the config name to use but config stanzas based on the task name are ignored:

```
[<taskname>]
foo = bar
```

@schettino72 said that this is not currently possible.

smarie:
Thanks @rbdixon , I added a review to your PR

That feature is very important for our use cases: indeed we need users to provide secret API keys. Of course we could do it in custom code, but it would make a lot of sense to have it in `doit`.

arichiardi:
Hi there, sorry if I am hijacking this, but is this feature only working for `python_action` - not for `cmd_action`?

I had to the following but maybe I am doing something wrong:

```python
 def image_action(docker_registry, docker_repository):
        image_tag = docker_image_qualified_name(docker_registry, docker_repository, namespace_tag)
        cmd = ["docker", "build", "--tag", image_tag, Path(".")]
        returncode = subprocess.call(cmd, shell=False)
        if returncode != 0:
            return doit.exceptions.TaskFailed(
                "Command failed: '%s' returned %s" % (cmd, returncode))

    return {
        "file_dep": [REACTOR_JAR_PATH, dockerfile_path],
        "actions": [image_action],
        "uptodate": [False],
        "params": [{"name": "docker_registry",
                    "long": "docker-registry",
                    "type": str,
                    "default": None,
                    "help": "The url to the docker registry."},
                   {"name": "docker_repository",
                    "long": "docker-repository",
                    "type": str,
                    "default": None,
                    "help": "The string representing the docker repository."}],
```

rbdixon:
Yes, I'd open a new issue or submit a patch if possible. A minimal test case is always useful. What you have above is probably close. Just prune out all of the details and try to make it as simple as possible. Mention me in the new issue and I can take a look in January.

PR #321, opened 2019-07-26
Specifying options (params) for subtasks

smarie:
A step towards #311  .

I managed to make my propoal work: i.e. a return statement after the generator in a group task will modify the params on the group task AND the ones in the subtasks.

So you can for example write:

```python
from doit.tools import title_with_actions

def task_mytask():
    yield {
            'name': 'mysubtask',
            'actions': ["echo myparam is %(myparam)s"],
            'title': title_with_actions,
            'verbosity': 2,
        }

    # new: the final return statement defines params
    # for the group task AND all yielded subtasks
    return {'params': [{'name': 'myparam',
                        'long': 'myparam',
                        'default': None}]}

```

Currently it only works if the parameter is passed to a **named** subtask through the CLI: for example `doit mytask:mysubtask --myparam hello` will give

```
.  mytask:mysubtask => Cmd: echo myparam is %(myparam)s
myparam is hello
```

(By the way the title is not updated, this is maybe another issue.)

However the parameter is not correctly propagated if I call the group task instead `doit mytask --myparam hello` yields

```
.  mytask:mysubtask => Cmd: echo myparam is %(myparam)s
myparam is None
```

Any idea why ?

I can not figure out from the documentation how to propagate a parameter from the CLI or from the configuration file to a task without naming the task (and in this case the subtask) explicitly. Any help would be appreciated so that I can finalize the proposal.

The PR still misses unit tests and conceptual validation of what is been done: as of now I copy the group task params to all subtasks.

Thanks!

### DIFF
--- a/doit/loader.py
+++ b/doit/loader.py
@@ -49,8 +49,9 @@ def flat_generator(gen, gen_doc=''):
                                             "return statement (to customize a"
                                             " group Task).")
                         else:
-                            # generate a new StopIteration by returning None
-                            return
+                            # break from this loop as this generator has nothing more
+                            # to provide
+                            break
                     else:
                         yield value, value_doc
             else:
@@ -334,6 +335,7 @@ def _generate_tasks_from_generator(func_name, gen_result, gen_doc):
                     group_task = tasks.get(func_name)
 
                     # override contents
+                    # TODO maybe move part of this code into a task.update() method for consistency
                     for k, v in extra_info.items():
                         # TODO should we support overriding other items?
                         if k in ('params',):
--- a/doit/loader.py
+++ b/doit/loader.py
@@ -43,7 +43,7 @@ def flat_generator(gen, gen_doc=''):
                         # make sure that the user did not make a mistake by
                         # using return in the wrong place
                         if e.value is not None:
-                            raise Exception("Invalid `return` statement in"
+                            raise Exception("Invalid `return` statement in "
                                             "nested generator. Only root-level"
                                             "generators can include a non-None"
                                             "return statement (to customize a"
--- a/doit/loader.py
+++ b/doit/loader.py
@@ -21,15 +21,40 @@ TASK_STRING = "task_"
 
 def flat_generator(gen, gen_doc=''):
     """return only values from generators
-    if any generator yields another generator it is recursively called
+    if any generator yields another generator it is recursively called.
+    `return` statements in the generators will be correctly propagated.
+    Note that only the root-level generator is allowed to return a
+    non-None value (to customize the group task)
     """
-    for item in gen:
-        if inspect.isgenerator(item):
-            item_doc = item.gi_code.co_consts[0]
-            for value, value_doc in flat_generator(item, item_doc):
-                yield value, value_doc
+    while True:
+        try:
+            item = next(gen)
+        except StopIteration as e:
+            # make sure we re-raise the contents!
+            return e.value
         else:
-            yield item, gen_doc
+            if inspect.isgenerator(item):
+                item_doc = item.gi_code.co_consts[0]
+                g = flat_generator(item, item_doc)
+                while True:
+                    try:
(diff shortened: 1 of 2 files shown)


smarie:
I added positive and negative tests.
The problem with Appveyor does not seem related to my commits (database teardown issue).

schettino72:
This design does not solve the case where argument values are used in the task-creator itself.
And I think the interface is completely non-intuitive. So a no go, sorry.

smarie:
> This design does not solve the case where argument values are used in the task-creator itself.

So do you mean what [`letsdoit`](https://bitbucket.org/takluyver/letsdoit) plugin is already doing ?
Then ok, that's another topic indeed.

But still I can not determine from the documentation (and I read it quite a number of times now) why the parameters 

 - can not flow from the `doit.cfg` config file to the task (I can put the `myparam` parameter in any section, it is not propagated)
 - can not flow from the commandline to the subtask when the subtask is not explicitly named on the command:

    - `doit --myparam hello` does not propagate it
    - `doit mytask --myparam hello` does not propagate it
    - `doit mytask:mysubtask --myparam hello` works

I'll try to reverse-engineer the code on monday to understand why and if it is a bug or a misunderstanding from my part.

schettino72:
>>   This design does not solve the case where argument values are used in the task-creator itself.
> 
> So do you mean what letsdoit plugin is already doing ?

I mean to be able to use argument value while creating the task dict. Not only on task execution (actions). Something like this

```
from doit import task_param

@task_param({'name':'mytarget', 'short':'t', 'default':'result.txt'})
def task_py_params(target):
        return {
            'name': name,
            'actions':['echo hello > %(targets)'],
            'targets': [mytarget]
        }
```

schettino72:
can not flow from the doit.cfg config file to the task (I can put the myparam parameter in any section, it is not propagated)

yes. this is a limitation: see #283

    can not flow from the commandline to the subtask when the subtask is not explicitly named on the command:

There is no such a thing of flow of parameters, but...

        doit --myparam hello does not propagate it

You can use ``get_var`` to pass global arguments from command line.

https://pydoit.org/task_args.html?highlight=get_var#command-line-variables-doit-get-var

        doit mytask --myparam hello does not propagate it

This should be achievable after #311 is implemented

rbdixon:
See PR #322 for an implementation for issue #283 to allow task parameters to be defined in the config file.

smarie:
@schettino72 I just read your message again, and it seems that therefore one feature is missing and not already present in the backlog: the capability to expose task parameters to the commandline even when several tasks are run (typically when the default task sequence is run, and the user only calls `doit`). 

Shall we open a ticket for this ? I have to admit that you have been closing tickets so fast in the past days that now I hesitate a bit ;)

schettino72:
@smarie have you read this? https://github.com/pydoit/doit/blob/master/.github/ISSUE_TEMPLATE/feature_request.md

If you follow your tickets will have a better chance of not being closed "so fast".

smarie:
I have, actually :) and I tried my best to follow it. Apparently that was not enough 😄 
Let me try once again then, if the above is not yet in the backlog or doc.

rbdixon:
Very short comment since this is closed... IF @schettino72 thought it was a good idea to have a group-level means to define subtask params THEN I'd suggest moving this from being `return`-based to using `yield` with `name: None` as shown in https://pydoit.org/tasks.html#avoiding-empty-sub-tasks. To me, though, the benefit of this feature is not enough to be included since it is simple enough to use Python to accomplish the same objective.

smarie:
Thanks @rbdixon for this suggestion ! 

Indeed this is interesting: if there is an existing mechanism to set common `doc`and `watch` attributes, then it could be used to set common `parameters`. In that case there is no need to rely on the `return` statement.

It is strange that I did not come across this mechanism in the code where the subtasks are generated... Maybe this is done in a later step.

PR #322, opened 2019-07-28
Add feature to specify task parameters in doit.cfg (issue #283)

rbdixon:
doit.cfg options take precedence over task definition default values
but with be replaced by task parameters passed on the command line.

TODO:

- [x] Implement `[task:arg_from_cfg]`
- [x] Rename `cfg_defaults` to `cfg_values`
- [x] Remove task config checks for `cfg_values` parameter
- [x] Use `.overwrite_defaults()` to apply `cfg_values`
- [x] Refactor `control.py` per comment
- [x] Tests
    - Ensure that doit.cfg specified task parameters are applied.
- [x] Documentation

Implements #283 

Demo:

```
$ cat dodo.py
def task_arg_from_cfg():
    def runit(x=1, y=1):
        print(f'python x={x}, y={y}')
        # if x==1: raise ValueError()

    return {
        'actions': [runit, 'echo shell x=%(x)s y=%(y)s'],
        'params': [{'default': 1, 'name': 'x', 'long': 'x'}, {'default': 2, 'name': 'y', 'long': 'y'}],
        'verbosity': 2,
}
```
```
$ cat doit.cfg
[task:arg_from_cfg]
    x = 1000
```
```
$ doit
.  arg_from_cfg
python x=1000, y=2
shell x=1000 y=2

```
```
$ doit run arg_from_cfg
.  arg_from_cfg
python x=1000, y=2
shell x=1000 y=2

```
```
$ doit run arg_from_cfg --x=123
.  arg_from_cfg
python x=123, y=2
shell x=123 y=2

```

### DIFF
--- a/doit/cmd_base.py
+++ b/doit/cmd_base.py
@@ -388,10 +388,11 @@ class NamespaceTaskLoader(TaskLoader2):
                                  cmd.execute_tasks)
 
         # Add task options from config, if present
-        for task in tasks:
-            task_stanza = 'task:' + task.name
-            if task_stanza in self.config:
-                task.cfg_values = self.config[task_stanza]
+        if self.config is not None:
+            for task in tasks:
+                task_stanza = 'task:' + task.name
+                if task_stanza in self.config:
+                    task.cfg_values = self.config[task_stanza]
 
         return tasks
 
--- a/doit/cmd_base.py
+++ b/doit/cmd_base.py
@@ -384,9 +384,16 @@ class NamespaceTaskLoader(TaskLoader2):
         return loader.load_doit_config(self.namespace)
 
     def load_tasks(self, cmd, pos_args):
-        return loader.load_tasks(self.namespace, self.cmd_names,
+        tasks = loader.load_tasks(self.namespace, self.cmd_names,
                                  cmd.execute_tasks)
 
+        # Add task options from config, if present
+        for task in tasks:
+            task_stanza = 'task:' + task.name
+            if task_stanza in self.config:
+                task.cfg_values = self.config[task_stanza]
+
+        return tasks
 
 class ModuleTaskLoader(NamespaceTaskLoader):
     """load tasks from a module/dictionary containing task generators
@@ -577,12 +584,6 @@ class DoitCmdBase(Command):
         # load tasks from new-style loader
         if not legacy_loader:
             self.task_list = self.loader.load_tasks(cmd=self, pos_args=args)
-
-        # Add task options from config, if present
-        for task in self.task_list:
-            task_stanza = 'task:' + task.name
-            if task_stanza in self.config:
-                task.cfg_values = self.config[task_stanza]
 
         # hack to pass parameter into _execute() calls that are not part
         # of command line options
--- a/doit/cmd_base.py
+++ b/doit/cmd_base.py
@@ -580,7 +580,9 @@ class DoitCmdBase(Command):
 
         # Add task options from config, if present
         for task in self.task_list:
-            task.cfg_defaults = self.config.get(task.name, {})
+            task_stanza = 'task:' + task.name
+            if task_stanza in self.config:
+                task.cfg_values = self.config[task_stanza]
 
         # hack to pass parameter into _execute() calls that are not part
         # of command line options
--- a/doit/cmd_base.py
+++ b/doit/cmd_base.py
@@ -578,6 +578,10 @@ class DoitCmdBase(Command):
         if not legacy_loader:
             self.task_list = self.loader.load_tasks(cmd=self, pos_args=args)
 
+        # Add task options from config, if present
+        for task in self.task_list:
+            task.cfg_defaults = self.config.get(task.name, {})
+
         # hack to pass parameter into _execute() calls that are not part
         # of command line options
         params['pos_args'] = args
--- a/doit/cmdparse.py
+++ b/doit/cmdparse.py
@@ -333,7 +333,7 @@ class CmdParse(object):
         return params, args
 
 
-    def parse(self, in_args, cfg_defaults={}):
+    def parse(self, in_args):
         """parse arguments into options(params) and positional arguments
 
         Also get values from shell ENV.
@@ -342,7 +342,6 @@ class CmdParse(object):
         an item for every option.
 
         @param in_args (list - string): typically sys.argv[1:]
-        @param cfg_defaults (dict): dict of values parsed from doit.cfg
         @return params, args
              params(dict): params contain the actual values from the options.
                            where the key is the name of the option.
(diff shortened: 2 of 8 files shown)


rbdixon:
I'm working on some test cases but I wanted to get this preliminary implementation out for review. Let me know what you think, please.

doit/cmdparse.py
```diff
@@ -333,7 +333,7 @@ def parse_only(self, in_args, params=None):
         return params, args
 
 
-    def parse(self, in_args):
+    def parse(self, in_args, cfg_defaults={}):
```
smarie: This name is a bit misleading, as it does not refer to the notion of "config file" and "defaults" would intuitively refer to a default value in the task definition, not in the config file.

I would suggest `cfg_from_file`, `cfg_from_doit_cfg_file`, `file_provided_cfg`, `cfgfile_args`, or similar.

rbdixon: Thanks, I think you are right. I'll tweak that.

doit/cmdparse.py
```diff
@@ -333,7 +333,7 @@ def parse_only(self, in_args, params=None):
         return params, args
 
 
-    def parse(self, in_args):
```
schettino72: It is NOT ok to make any changes on `cmdparse.py` module to fix this issue.
`cmdparse` already has support to handle different sources of values (although not explicitly).
You should use `overwrite_defaults()`, see other comment...

doit/control.py
```diff
@@ -158,7 +158,7 @@ def add_filtered_task(seq, f_name):
                 the_task = self.tasks[f_name]
                 # remaining items are other tasks not positional options
                 taskcmd = TaskParse([CmdOption(opt) for opt in the_task.params])
```
schettino72: This part here has some code duplication with `Task.init_options`.
My bad, we should refactor this so the actual change to fix this issue occurs in only one place.

doit/task.py
```diff
@@ -189,6 +191,7 @@ def __init__(self, name, actions, file_dep=(), targets=(),
         self.check_attr(name, 'getargs', getargs, self.valid_attr['getargs'])
         self.check_attr(name, 'title', title, self.valid_attr['title'])
         self.check_attr(name, 'watch', watch, self.valid_attr['watch'])
+        self.check_attr(name, 'cfg_defaults', cfg_defaults, self.valid_attr['cfg_defaults'])
```
schettino72: All these initializations and check are for values provided by users in `dodo.py` files when creating tasks.
That is not the case for these config values. So no need to change `Task.__init__`, just initialize the value to `None`.

doit/task.py
```diff
@@ -367,7 +371,7 @@ def init_options(self):
         if self.options is None:
             taskcmd = TaskParse([CmdOption(opt) for opt in self.params])
             # ignore positional parameters
-            self.options = taskcmd.parse('')[0]
```
schettino72: here is the key point. Before calling `taskcmd.parse` we should call:

```
taskcmd.overwrite_defaults(self.cfg_defaults)
```

doit/task.py
```diff
@@ -230,6 +233,7 @@ def __init__(self, name, actions, file_dep=(), targets=(),
         self.values = {}
         self.verbosity = verbosity
         self.custom_title = title
+        self.cfg_defaults = cfg_defaults
```
schettino72: I think `cfg_values` would be a better name.

doit/cmd_base.py
```diff
@@ -578,6 +578,10 @@ def execute(self, params, args):
         if not legacy_loader:
             self.task_list = self.loader.load_tasks(cmd=self, pos_args=args)
 
+        # Add task options from config, if present
+        for task in self.task_list:
+            task.cfg_defaults = self.config.get(task.name, {})
```
schettino72: It bothers me a bit that we might be calling `config.get` for every task when rarely it will contain a value...

rbdixon:
This is very helpful and I'll rework the patch. I really appreciate the thoughtful review.

doit/cmd_base.py
rbdixon: It is roughly 30% faster to do an `if task.name in self.config: ...`. What the actual execution time benefit would be depends on how many tasks are defined.

Benchmark:

https://gist.github.com/rbdixon/797d0478f9716c4123e401a1187552e1

```
Number of tasks: 1326
Speedup -30.40%
Baseline duration: 0.25ms
Alternative duration: 0.18ms
```

Might be hard for a user to tell the difference in execution time either way.

doit/task.py
rbdixon: Fixed. I've removed the initializations and checks.

doit/cmdparse.py
rbdixon: I've reverted modifications to this file.

doit/task.py
rbdixon: I've fixed this in both locations. Thanks... I failed to notice `.overwrite_defaults()` was available. Much clearer.

doit/task.py
rbdixon: Changed.

doit/control.py
rbdixon: If you have a preferred place to centralize this code I can attempt. What would you recommend?

schettino72: I guess I would create a new method in `Task` class. `Task.set_options` or maybe just re-use `Task.init_options`.

doit/task.py
```diff
@@ -165,7 +164,7 @@ def __init__(self, name, actions, file_dep=(), targets=(),
                  subtask_of=None, has_subtask=False,
                  doc=None, params=(), pos_arg=None,
                  verbosity=None, title=None, getargs=None,
-                 watch=(), loader=None, cfg_defaults=dict()):
+                 watch=(), loader=None, cfg_values=None):
```
schettino72: better remove from `__init__` parameters. It is not expected to set this value on Task creation.

rbdixon: Fixed.

doit/control.py
rbdixon: I've implemented this by tweaking `Task.init_options()`:

```python
    def init_options(self, args=None):
        """Put default values on options.

        This can be called with optional command line task arguments.

        Return value: unparsed command line task arguments.
        """
```

If `args` is provided the command line arguments are parsed and the return value is the list of remaining unparsed arguments. If `args` is not specified the return value is `None`.

rbdixon:
I've implemented tests, documentation, and all changes discussed in this thread. I'm using this branch in my daily work with our own internal tools and nikola. So far, so good.

 I believe this is ready to be merged.

arichiardi:
This is a great feature, just ran into this and want to say thank you and bump it :smiley:

schettino72:
merged: https://github.com/pydoit/doit/commit/f0303f2390368fd10d63e5b9bbd05d70dc4dfe34

thanks. and sorry for delay.

PR #326, opened 2019-07-30
New `BaseAction.to_str` method used in particular in `title_with_actions`.

### LINKED ISSUE #325: Task title created by `title_with_actions` does not replace parameter and options values.
Consider the following `dodo.py` file:

```python
DOIT_CONFIG = {'default_tasks': ['mytask']}


from doit.tools import title_with_actions


def task_mytask():
    return {
            # 'name': 'mysubtask',
            'params': [{
                'name': 'myparam',
                'short': 'p',
                'long': 'myparam',
                'default': 'param_default',
            }],
            'actions': ["echo myparam is %(myparam)s"],
            'title': title_with_actions,
            'verbosity': 2,
        }
```

Executing it yields:

```shell
>>> doit
.  mytask => Cmd: echo myparam is %(myparam)s
myparam is param_default
```

You can notice that the title created for the action is not correct: one would expect to see
`Cmd: echo myparam is param_default`, in other words one would expect that this title, since it is created after task instantiation, actually represents what will be executed.

I see two ways to fix this:

 - fixing only the `title_with_actions` method. This would force us to write an ugly if/elif statement in it to support all kind of `BaseAction` subtypes.
 - fixing the `__str__` in all `BaseAction` subtypes directly. For example for `CmdAction`:

```python
    def __str__(self):
        # try to replace all options in the various commands
        try:
            _action = self.expand_action()
        except Exception:
            _action = self._action

        return "Cmd: %s" % _action
```

smarie:
New `BaseAction.to_str` method so that all actions can provide a string representation with or without expanding the options inside. 

This method is used by helper function `title_with_actions`. 

Added corresponding tests. 

Fixes #325

Note that this PR also improves slightly the string representation of python functions, as a side effect.

Still todo: documentation.

### DIFF
--- a/doit/action.py
+++ b/doit/action.py
@@ -94,6 +94,19 @@ class BaseAction(object):
                 kwargs[key] = opt_args[key]
         return kwargs
 
+    def to_str(self, expand_options):
+        """
+        Should return a string representation of this action,
+        with options taken into account if `expand_options` is `True`.
+
+        Subclasses should override this to expand the options properly.
+
+        :param expand_options: a boolean indicating if the string representation
+            should contain the actual options used by the action in current
+            execution context.
+        :return:
+        """
+        return str(self)
 
 
 class CmdAction(BaseAction):
@@ -310,7 +323,15 @@ class CmdAction(BaseAction):
             return self.action.format(**subs_dict) % subs_dict
 
     def __str__(self):
-        return "Cmd: %s" % self._action
+        return self.to_str(expand_options=False)
+
+    def to_str(self, expand_options):
+        if expand_options:
+            a = self.expand_action()
+        else:
+            a = self._action
+
+        return "Cmd: %s" % a
 
     def __repr__(self):
         return "<CmdAction: '%s'>" % str(self._action)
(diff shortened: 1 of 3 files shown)


schettino72:
Maybe I was not clear enough...
A change in `title_with_actions` to support expanding task parameters would be good.
But I did NOT agree to any changes in `actions` module.

schettino72:
`actions` can now use the more modern string formatting with `{}` instead of `%`.
Usage of `%` in action strings will be deprecated... 
Moving forward would be better to support `{}` instead of `%`.

smarie:
I have a hard time to understand your last comment: I thought that `self.expand_action()` in `CmdAction` was doing the replacement, taking into account all the formatting modes that you refer to  ? 

In this PR as you can see 

 - in `CmdAction` I call `self.expand_action()` so whatever logic is here, it is reused
 - in `PythonAction` there was nothing here but it is simpler: I just display the arguments

If you prefer to put a big if/then in `title_with_actions` to support all `BaseAction` subtypes it is your call, but IMHO it is not the most maintainable choice in the long run especially if new action types emerge.

schettino72:
Refusing this. Sorry. I agreed only with changes on `title_with_actions`.

smarie:
No worries @schettino72 , thanks for the feebdack. By the way I now work using [doit-api](https://smarie.github.io/python-doit-api/), do not hesitate to have a quick look ! It completes `doit` so as to make it touch a broader audience: more users for you :)

PR #330, opened 2019-08-24
Add decorator to mark a class method as a task generator

zspiderx:
1. `@TaskGenerator.doit_task` can be used to mark a class method as a task generator
2. If a class is a subclass of `TaskGenerator`, `doit_tasks` can be used on the class to return a dict of all the task generator.


schettino72:
cool. doit was designed to be easy to create helpers like theis...
but these kind of helpers do not belong to doit core.

PR #340, opened 2019-12-08
.travis.yml: Add Python 3.8

jayvdb:


.travis.yml
```diff
@@ -3,13 +3,10 @@ python:
...
     - "3.5"
     - "3.6"
+    - "3.7"
+    - "3.8"
     - "pypy3.5"
 
-matrix:
-  include:
-    - python: 3.7
-      dist: xenial
```
hartwork, line 11: In case someone else wonders:
This change seems cosmetic in nature, as xenial seems to be [the default as of today](https://docs.travis-ci.com/user/reference/xenial/).

PR #347, opened 2020-01-21
Drop support for Python 3.4

hartwork:
Because of:
> DEPRECATION: Python 3.4 support has been deprecated. pip 19.1 will be the last one supporting it. Please upgrade your Python as Python 3.4 won't be maintained after March 2019 (cf PEP 429).

See https://travis-ci.org/pydoit/doit/jobs/640146558#L213


hartwork:
Obsolete, closing.

Issue #341, opened 2019-12-09
Test failure on Python 3.8


jayvdb:
```
TestJobTask.test_not_picklable_raises_InvalidTask _______________

self = <tests.test_runner.TestJobTask object at 0x7f3ec7fb9040>

    @pytest.mark.xfail('PLAT_IMPL == "PyPy"')  # pypy can handle it :)

    def test_not_picklable_raises_InvalidTask(self):

        # create a large enough recursive obj so pickle fails

        d1 = {}

        last = d1

        for x in range(400):

            dn = {'p': last}

            last = dn

        d1['p'] = last

    

        def non_top_function(): pass

        t1 = Task('t1', [non_top_function, (d1,)])

>       pytest.raises(InvalidTask, runner.JobTask, t1)

E       Failed: DID NOT RAISE <class 'doit.exceptions.InvalidTask'>
```
c.f. https://github.com/pydoit/doit/pull/340 and https://travis-ci.org/pydoit/doit/builds/622302741

PR #350, opened 2020-02-02
Py3.8: Replace recursive knot with explicitly unpicklable object

stefanor:
Python 3.8 was able to pickle the previously unpicklable. Instead of relying on limits, let's raise an explicit error.

Fixes: #341


schettino72:
thanks

PR #360, opened 2020-05-16
doc: Add :abbr: roles

xordspar0:


schettino72:
Makes no much sense to put options from different sub-commands together...

PR #374, opened 2020-10-31
Add TOML config loading, autodetect pyproject.toml

### LINKED ISSUE #373: Read DOIT_CONFIG from (pyproject).toml
Thanks for doit! Per the instructions on the template, this has a corresponding thread on the [mailing list](https://groups.google.com/forum/#!topic/python-doit/NOZm_4EWYD8).

## Elevator Pitch

Allow `doit`, if [`toml`](https://github.com/uiri/toml) is installed, to look for `DOIT_CONFIG` in [`pyproject.toml`](https://www.python.org/dev/peps/pep-0518/) by default, or any given [TOML](https://toml.io/en/v1.0.0-rc.3) file.

## Examples

TOML is quite similar, but a bit more flexible, than the `ConfigParser` syntax. A `pyproject.toml` most like the  [test `doit.cfg`](https://github.com/pydoit/doit/blob/master/tests/sample.cfg):

```toml
[build-system]
requires = ["setuptools >= 40.6.0", "wheel"]
build-backend = "setuptools.build_meta"

[tools.doit.DOIT_CONFIG.GLOBAL]
optx = 6
opty = 7

[tools.doit.DOIT_CONFIG.COMMAND]
foo = "tests.sample_plugin:MyCmd"
```

Though users may prefer alternatives, with more commas:

```toml
[build-system]
requires = ["setuptools >= 40.6.0", "wheel"]
build-backend = "setuptools.build_meta"

[tools.doit.DOIT_CONFIG]
GLOBAL = {
  optx = 6,
  opty = 7,
}
COMMAND = {
  foo = "tests.sample_plugin:MyCmd"
}
```

## Motivation

The python tool community is _slowly_ making progress out of dotfile hell by standardizing on `pyproject.toml`. `black`, `pytest`, `tox` (to an extent, and not yet compatible with `pytest` graah) and others have already carved out their own respective namespace under `tools`.

Additional tools like `dephell` and `flakehell` make managing this content reasonable, and integrating tools that don't yet support `pyproject.toml`.

## Technical Concerns

`pyproject.toml` was originally only intended for one heading, `[build-system]`, for allowing isolated build environments, and avoiding the chicken-and-egg of `setup_requires` in `setup.py` and `setup.cfg`. However, once it became clear that this file wasn't going anywhere, the official stance softened to allow any number of extra headings under the `tool` key. The [current explanation](https://laptrinhx.com/brett-cannon-what-the-heck-is-pyproject-toml-3531099003/) provides some insight. Of note, a project can't yet _entirely_ replace `setup.py` for a "bog standard" `setuptools` build, but this is in-progress on [PEP621](https://www.python.org/dev/peps/pep-0621/).

The TOML language itself is not without quirks, and the underlying spec is not yet `1.0`, but is already widely used for Officially Important Files in the julia and rust, and now python, packaging communities, and is unlikely to change in a backwards-incompatible way.

A TOML parser is not part of the python standard library, but the python `toml` package supports the `1.0.0rc*` spec, and is already relied upon by a number of tools (e.g. `black`, `flit`).

## Proposal

I'm happy to work up a PR that:

/blob/f61d4ed9d587575fd20ee5ece1674ddc3088ee2a/src/_pytest/config/findpaths.py#L65-L79))
- Adds an `extras` for `toml` to `setup.py`
  - With a reasonable version pin
- Adds `pyproject.toml` to the default list of config files (right after `doit.cfg`, which should still always "win")
- Pre-screens `pyproject.toml` (or whichever `.toml` the user might specify) for the magic `DOIT_CONFIG` string
  - If found...
    - try to import `toml` (a la [`pytest`](https://github.com/pytest-dev/pytest 
      - if this fails, show a warning
    - otherwise parses and extracts `parsed["tools"]["doit"]["DOIT_CONFIG"][*]`
      - shims into the final `ConfigParser` output
- Tests!
- Docs!

## Alternatives
- make `toml` a hard requirement (a la [black](https://github.com/psf/black/blob/master/setup.py#L73))
  - :+1: don't have to specify an install
  - :-1: extra, unused dependency for many projects
- vendor `toml`
  - :+1: don't have to specify an install
  - :-1: have to manage version, etc.
  - :-1: vendoring is generally annoying from a downstream packaging perspective
    - _disclaimer: I'm one of the [conda-forge/doit-feedstock](https://github.com/conda-forge/doit-feedstock) maintainers_
- create a first- or third-party plugin, e.g. `doit-config-loader-toml`
  - :-1: potential version mismatches
- do-nothing: don't support `pyproject.toml` at all
  - :+1: no work!
  - :+1: no extra dependencies
  - :-1: still have a random `doit.cfg` rolling around in my repo

## Follow-on 

- with `toml` available, it may make sense to be able to load _simple_ tasks from `pyproject.toml` in a predictable manner
  - especially those that are configured _in_ `pyproject.toml` 
  - these could be intermixed with complex tasks (e.g. advanced `yield`s, etc) from `dodo.py` or wherever
  - perhaps rather than the approach on the [yaml loader](https://github.com/pydoit/doit-yaml), provide something that, initially, more closely matches the raw strings, etc. that can instantiate a task
    - there are challenges with things like file globs

```toml
[build-system]
requires = ["setuptools >= 40.6.0", "wheel"]
build-backend = "setuptools.build_meta"

[tools.black]
line-length = 80

[tool.flakehell]
max_line_length = 90
show_source = true

[tool.flakehell.plugins]
pyflakes = ["+*", "-F401"]

[tools.doit.DOIT_CONFIG.GLOBAL]
backend = "sqlite3"
verbosity = 2
par_type = "thread"

# this uses the "nested tables" approach, allowing for groups
[[tools.doit.tasks.lint]]  
name = "black"
doc = "format python code"
actions = ["black > build/black.ok"]
targets = ["build/black.ok"]
file_dep = ["foo.py"]              
# this is a whole other story, or perhaps, with some more opinions...
# file_dep = [{ glob = "**/*.py" }]

[[tools.doit.tasks.lint]]
name = "flakehell"
doc = "check for python flake"
actions = ["flakehell > build/flakehell.ok"]
targets = ["build/flakehell.ok"]
file_dep = ["foo.py"]
```

- such a spec could be formally encoded with, e.g. a [JSON schema](https://tools.ietf.org/html/draft-handrews-json-schema-01), and reused for other well-behaved declarative formats.

If you made it this far, thanks for reading!

bollwyvl:
- fixes #373

This adds TOML support, with detection of `pyproject.toml`.

While I like the suggestion on the parent issue of not using SHOUTY keys, I've tried to make it look as much like the existing config format as is reasonable, and we can iterate from there, but figured being able to comment on (maybe broken) code is better than more issue comment traffic.

TODO:
- [x] ~~there are still some ordering issues... right now, values from `doit.cfg` would be overwritten by `pyproject.toml`: need to look into the expected behavior from ConfigParser.~~
  - UPDATE: it is indeed FIFO... the merging is still quite naive, e.g. only of the top-level sections, but matches existing behavior
- [x] tests
  - will it toml?
  - will it pyproject.toml?
- [x] docs


doit/doit_cmd.py
```diff
@@ -52,18 +52,28 @@ class DoitMain(object):
                  Strace, TabCompletion, ResetDep)
 
     def __init__(self, task_loader=None,
-                 config_filenames='doit.cfg',
+                 config_filenames=['pyproject.toml', 'doit.cfg'],
```
bollwyvl, line 55: This is to align with the existing behavior of `ConfigParser.read`, which appears to be FIFO... I verified it, as i didn't see it clearly [documented](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read). but makes sense.

doit/doit_cmd.py
```diff
@@ -52,18 +52,28 @@ class DoitMain(object):
...
                 self.config[section].update(items)
-        ini_config = self.load_config_ini(config_filenames)
-        for section in ini_config.sections():
-            self.config[section].update(ini_config[section].items())
+
+        if isinstance(config_filenames, str):
+            config_filenames = [config_filenames]
+
+        for config_filename in config_filenames:
+            if str(config_filename).lower().endswith('.toml'):
```
bollwyvl, line 69: casting to a string, here, as it can be a `Path` (python 3.6.1+) or `bytes` (python 3.7).

bollwyvl:
@schettino72 any thoughts here? happy to help hammer out whatever else is missing!

doc/configuration.rst
```diff
@@ -34,6 +32,20 @@ it is processed. It supports 3 kind of sections:
 - a section for each command
 - a section for each plugin category
 
+.. note::
```
schettino72, line 35: We are adding first-class support to TOML... so it deserves its own section on docs. Not just a "note".

I would rather "duplicate" the docs than force users to read one and figure out how to translate it...

Also ok to put TOML first and say it is the preferred format (because it might include extra features).

schettino72:
> While I like the suggestion on the parent issue of not using SHOUTY keys, I've tried to make it look as much like the existing config format as is reasonable,

This is not a goal. The goal is to have as idiomatic TOML as possible. I understand you want to keep the code changes to minimum, but we could internally convert TOML format to what whatever the current code is expecting.

bollwyvl:
Great! I'll turn these comments around asap. I'm somewhat rst-illiterate so
the docs will be best-effort!

bollwyvl:
I've updated to more closely match the proposed style in https://github.com/pydoit/doit/issues/373#issuecomment-719164925. Turns out TOML is rather picky about whitespace around `{}` so some of the syntax aren't _quiet_ doable (though maybe I need to just learn more).

The translation stuff is multi-step, and perhaps a bit verbose, but seems fine.

I did a sweep of other docs spots, and mostly just duplicated the cfg stuff, as suggested, which didn't seem unreasonable. The multiline strings are nice for, e.g. the `auto` callbacks.

Think we're getting closer!

bollwyvl:
@schettino72 any thoughts? Happy to do whatever it takes to bring this home!

bollwyvl:
I've removed the deprecated `pytoml`, as the last in-the-wild user of it (`flit`) has dropped it.

bollwyvl:
All :heavy_check_mark: again :tada: !

bollwyvl:
@schettino72 any feedback here?

schettino72:
@bollwyvl 

Last year I was pretty bad at managing my "open-source" time. This year I am targeting 2 hours a week. Note doit is not my only open-source project...

There are still quite a few items before this PR. 
If you feel like making a financial contribution I would schedule to work on it on my standard working hours.

Regards

bollwyvl:
No worries there, and do appreciate the timely feedback.

bollwyvl:
I might add, regarding review workload: if there's are additional automation measures that would help reduce your review load, I'm glad to add anything desired to the workflows.

doc/configuration.rst
```diff
@@ -18,12 +18,94 @@ The name can be seem from ``doit help`` output::
...
+While mostly similar, `TOML <https://toml.io>`_ differs from the INI format
+in a few ways:
+
+- all strings must be quoted with `'` or `"`
+- triple-quoted strings may contain new line characters (`\n`) and quotes
+- must be saved as UTF-8
+- integers and floating point numbers can be written without quotes
+- boolean values can be written unquoted and lower-cased, as `true` and `false`
+
+Unlike "plain" TOML, `doit` will parse pythonic strings into their correct types,
```
schettino72, line 50: Why? I think this is in anti-feature. And I could not see where this comes from, all python libs work like this?

doc/configuration.rst
```diff
@@ -18,12 +18,94 @@ The name can be seem from ``doit help`` output::
...
+e.g. `"True"`, `"False"`, `"3"`, but using "native" TOML types may be preferable.
+
+
+tool.doit
+^^^^^^^^^
+
+The `tool.doit` section may contain command line options that will be used
```
schettino72, line 57: I guess mentioned this before...

Why not just `doit` instead of `tool.doit`? For me it seems unnecessary verbose.

doit/doit_cmd.py
schettino72: Not sure what you mean by FIFO. which file has priority, first or last?

doc/configuration.rst
```diff
@@ -18,12 +18,94 @@ The name can be seem from ``doit help`` output::
...
+
+To configure options for a specific task, use a section with
+the task name under `tool.doit.tasks`:
+
+.. code-block:: toml
+
+   [tool.doit.tasks.make_cookies]
+   cookie_type = "chocolate"
+   temp = "375F"
+   duration = 12
```
schettino72, line 103: You mentioned about being able to define tasks in TOML, right?

Have you thought about what the name would be? I would thing `doit.tasks`. So maybe get a different name for setting task parameters `doit.params.<task_name>`, what do you think?

doc/extending.rst
```diff
@@ -128,13 +128,22 @@ config plugin
...
-located at the module `my_plugins.py`::
+Example of command plugin configured in `pyproject.toml`, implemented in
+the *class* `FooCmd`, located at the module `my_plugins.py`:
 
- [COMMAND]
- foo = my_plugins:FooCmd
+.. code-block:: toml
+
+   [tools.doit.plugins]
+   COMMAND = { foo = "my_plugins:FooCmd" }
```
schettino72, line 139: what about? 

```
[doit.plugins.foo]
kind = "command"
obj = "my_plugins:FooCmd"
```

Better name for "obj" ?

doc/configuration.rst
bollwyvl: This is a hangover from being able to reuse the existing `.ini` parsing logic, which only speaks string.

doc/configuration.rst
bollwyvl: This is specified by [PEP 518](https://www.python.org/dev/peps/pep-0518/#tool-table).

doit/doit_cmd.py
bollwyvl: Last-read wins. It seems to be the convention used by [isort](https://pycqa.github.io/isort/docs/configuration/config_files/), [pytest](https://docs.pytest.org/en/6.2.x/customize.html#configuration-file-formats), while `black` _only_ supports `pyproject.toml`.

schettino72 (commented):
Sorry for delay. Here is another round of questions...

doc/configuration.rst
schettino72: oh. i see... I think that is not acceptable.

doc/configuration.rst
schettino72: I see it is their recommendation... although I dont really understand/agree why limit to only 2 top tables.

So another question. Why put `doit` configuration on `pyproject.toml` instead of going with out own file `doit.toml`?

doc/extending.rst
bollwyvl: yeah, naming things is hard. If it was purely data, the more verbose form would feel good, but i think in this case spareness wins.

I believe this could also be written

```toml
[doit.plugins.COMMAND]
foo = "my_plugins:FooCmd"
```

(have updated docs to reflect both)

doc/configuration.rst
bollwyvl: >  Why put doit configuration on pyproject.toml 

Indeed, that has a solid question, and mypy has fought back viciously, for some reason. But my feeling is that the more tools I can cram into a single, consensus-driven file, the better I'll be able to automate the creation (through schema, etc), updating, etc. vs code.

> doit.toml?

A fine idea as well, and could be added to the list. I'd imagine _this_ one _wouldn't_ have the `tool.doit` prefix.

doc/configuration.rst
bollwyvl: Welp, that `config` object (wherever it gets read from or built dynamically) gets handed around throughout the entire running application, it appears, and (only) the `cmd_options` know how to do the marshaling. It seems like changing (or even detecting the use of) that behavior would require deprecation of the ini format altogether, and a formal data schema throughout all these things... sounds very 1.0ish.

schettino72:
Hey, I finally have merged this. It would be great if you could do one more pass reviewing this.

I went with lowercase version like:

```
[tool.doit.plugins.command]
foo = "tests.sample_plugin:MyCmd"
```

I also added support to loading from TOML file without `tool.doit`.
Just realized command line there is no way specify a config file, so it is only available through API call for now.

Thanks

bollwyvl:
Sounds great, thanks!

PR #379, opened 2020-12-08
Add github actions

### LINKED ISSUE #378: Add GitHub Actions?
Travis CI will become [all but unusable](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) within a month or two, and appveyor is... appveyor.

I'd be happy to do a PR (and test drive against my own `master`) to demonstrate GitHub Actions as a path forward.

The GitHub Actions YAML is not a huge change from either of the existing platforms, integrate nicely with the PR process, and actually _remove_ files from the top-level directory, which is always nice. There are some UX/DX issues I don't like, but _generally_ it's fewer clicks to get to what has broken than with third-party providers. _At present_, for open source projects, the minute allotment is unlimited and 20 machines can be running at once.

Unlike travis/appveyor, there's built-in artifact archiving, which is nice for things like coverage reports, built docs site previews, etc.

## Alternatives
- Azure Pipelines: needs another account, but really quite good, and have lots of reporting features
- Drone: needs another account, etc.
- Circl-CI: needs another account, etc.

bollwyvl:
- fixes #378

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,52 @@
+name: CI
+
+on:
+  push:
+    branches: [master, test]
+  pull_request:
+    branches: [master, test]
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu, windows, macos]
+        python-version: ['3.5', '3.6', '3.7', '3.8', '3.9', 'pypy3']
+        include:
+          # https://github.com/pydoit/doit/issues/372
+          - os: macos
+            pytest-args: -k 'not(cmd_auto or TestFileWatcher)'
+          - os: macos
+            python-version: pypy3
+            pytest-args: -k 'not(cmd_auto or TestFileWatcher or remove_all or ForgetAll)'
+        exclude:
+          - os: windows
+            python-version: pypy3
+    steps:
+      - if: ${{ matrix.os == 'ubuntu' }}
+        run: sudo apt-get install strace
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,52 @@
...
+            ${{ runner.os }}-pip-
+      - run: pip install . -r dev_requirements.txt python-coveralls
+      - run: pip freeze
+      - run: pip check
+      - run: doit pyflakes
+      - run: py.test -vv ${{ matrix.pytest-args }}
+      - if: ${{ matrix.os == 'ubuntu' && matrix.python-version == '3.8' }}
+        run: |
+          doit coverage
+          coveralls || echo "TODO: remove when run on a repo that is set up"
```
schettino72, line 52: What this means? "which kind of repo setup"?

bollwyvl: I hadn't set up codecov in a while, since the bad old days when a token was needed. I assume it's just working now?

schettino72: Ok understood. No, it is not working:

```
INFO:coveralls:422
INFO:coveralls:{"message":"Couldn't find a repository matching this job.","error":true}
TODO: remove when run on a repo that is set up
```

I dont understand this error message. It seems to work well on travis... Any idea why ?

bollwyvl: Yeah, there's probably a token or something. And looking at it, of course, it's `coveralls`. It looks it does need a token in secrets, which I don't like to bother with until i absolutely must.

I have been using `codecov` recently, which doesn't require a secret for public repos, and only generates _one_ PR comment, which it updates, which I find more convenient... will do a follow-up PR with a demo.

bollwyvl: PR up https://github.com/pydoit/doit/pull/384 with lots of :heavy_check_mark:

PR #382, opened 2021-02-23
Use uniform date format for TaskResult to_dict()

kohtoa15:
Standard datetime to string conversion leaves open the possibility that in case of µs = 0, no decimals are displayed, which could break conversion back to datetime based on standard format.


kohtoa15:
@schettino72

PR #386, opened 2021-03-18
Update cmd_auto.py

geo7:
fix typo


doit/cmd_auto.py
```diff
@@ -1,4 +1,4 @@
-"""starts a long-running process that whatches the file system and
```
schettino72, line 1: what? :grin:

schettino72:
Thanks

Issue #387, opened 2021-03-30
No output from failed tasks when multiple processes are used


alexandre-allard-scality:
**Describe the bug**

When running doit with `--process 2` (or more), if an action fails, `err` and `out` attributes of the action are always empty in the reporter when captured.
However it works well if not using multiple processes or if the task succeeds.

Here is a simple `dodo.py` file to reproduce:
```
import doit


DOIT_CONFIG = {
    "reporter": doit.reporter.JsonReporter,
}

def task__test_failure():
    return {
        "actions": [doit.action.CmdAction("mkdir")],
    }
```

Which gives, without `--process 2`:
```
{
  "tasks": [
    {
      "name": "_test_failure",
      "result": "fail",
      "out": "",
      "err": "mkdir: missing operand\nTry 'mkdir --help' for more information.\n",
      "error": "Command failed: 'mkdir' returned 1\n",
      "started": "2021-03-30 08:11:03.319472",
      "elapsed": 0.0035848617553710938
    }
  ],
  "out": "",
  "err": "mkdir: missing operand\nTry 'mkdir --help' for more information.\n"
}
```

And with it:
```
{
  "tasks": [
    {
      "name": "_test_failure",
      "result": "fail",
      "out": "",
      "err": "",
      "error": "Command failed: 'mkdir' returned 1\n",
      "started": "2021-03-30 08:11:12.858163",
      "elapsed": 0.0039060115814208984
    }
  ],
  "out": "",
  "err": ""
}
```

**Environment**
1. OS: Ubuntu 20.04
2. python version: 3.8.5
3. doit version: 0.32.0

alexandre-allard-scality:
Another issue, not directly related to this one but it is the same part in the code, when the task fails, attributes are not updated with the pickled ones.
Meaning that a failed task is always seen as not executed.

PR #388, opened 2021-03-31
Pass pickled attributes, err and out on failure

alexandre-allard-scality:
In multi processes context, the pickled attributes,
stderr and stdout were not returned in case of
TaskError or TaskFailed, thus making it hard to see
what gone wrong when a command exits in error.

Refs: #387

### DIFF
--- a/doit/runner.py
+++ b/doit/runner.py
@@ -435,16 +435,12 @@ class MRunner(Runner):
 
     def _process_result(self, node, task, result):
         """process result received from sub-process"""
-        if 'failure' in result:
-            catched_excp = result['failure']
-        else:
-            # success set values taken from subprocess result
-            catched_excp = None
-            task.update_from_pickle(result['task'])
-            for action, output in zip(task.actions, result['out']):
-                action.out = output
-            for action, output in zip(task.actions, result['err']):
-                action.err = output
+        catched_excp = result.get('failure')
+        task.update_from_pickle(result['task'])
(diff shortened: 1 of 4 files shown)


doit/runner.py
```diff
@@ -543,13 +539,12 @@ def execute_task_subprocess(self, job_q, result_q, reporter_class):
 
                 result = {'name': task.name}
                 t_result = self.execute_task(task)
-
-                if t_result is None:
-                    result['task'] = task.pickle_safe_dict()
-                    result['out'] = [a.out for a in task.actions]
-                    result['err'] = [a.err for a in task.actions]
-                else:
+                if t_result:
```
schettino72, line 542: `t_result` is actually a misleading variable name. It only has a value if there was a failure.
What about changing it to `t_failure` ?

schettino72 (changes requested):
Thanks. Could you please:
- add tests
- add an entry on CHANGES file
- add your name in AUTHORS file (only if you feel like, not required)

doit/runner.py
alexandre-allard-scality: Totally agree, I'll change it.

alexandre-allard-scality: Done

schettino72:
merged. thanks
Your test was not actually testing a task failure, so i added another test.
I used "git merge --squash", seems didnt keep you as author (sorry).

alexandre-allard-scality:
np, thanks for the merge.

PR #389, opened 2021-04-05
Update version on home page

boidolr:
Having a release from 2019 makes it seem less active than the project is.
Python version range is updated to what was supported according to the build jobs at the time of the release.


schettino72:
Thanks. Also added a note on README so that it doesnt help again.
https://github.com/pydoit/doit/commit/9efe141a5dc96d4912143561695af7fc4a076490

I will actually update the site another day...

Issue #390, opened 2021-04-08
Ability to print action before executing it.


Jasha10:
Feature request: ability to print some data about each action before it gets executed.

## Motivation:

If my task has multiple actions (and verbosity==2), I can't tell which output came from which action (assuming that the actions all succeed). Some way to tell which output comes from which action would be great. Printing e.g. the shell command that will be run, before running it, would be a good way to achieve this.

If you write `echo 123` as part of a recipe in a Makefile (using GNU make), the following is printed:
```bash
+ echo 123
123
```
This is great for debugging purposes, so that you will know what command has produced the output you are seeing. The `+` symbol prefix makes it easy to tell what is the command, and what is the output from the command.

## Possible API design:

Perhaps this could be verbosity level 3?

For `CmdAction`, it should be easy to print the action that will be executed by `Popen`, just before it gets executed.

For `PythonAction`, it is not as clear what should be the output. Perhaps a `str` representation of the python-action and it's args and kwargs...

## Describe alternatives you've considered
I have tried using the `title` keyword for the task, with a custom title printer:
```python
# dodo.py
import doit

def show_cmd(task):
    result = "EXECUTING " + task.name
    for action in task.actions:
        if isinstance(action, doit.action.CmdAction):
            result += "\n+ " + action.expand_action()
        if isinstance(action, doit.action.PythonAction):
            result += f"\n+ {action.py_callable.__name__}(*{action.args}, **{action._prepare_kwargs()})"
    return result

def py_action(a=1, b=3):
    print(f"in py_action: {a=}, {b=}")

def task_custom_display():
    return {"actions": ["echo abc efg", (py_action, (2,), {"b": 4})], "title": show_cmd}
```
However, when I run `doit --verbosity 2` on this `dodo.py` file, the results are out of order:
```bash
$ doit --verbosity 2
.  EXECUTING custom_display
+ echo abc efg
+ py_action(*(2,), **{'b': 4})
abc efg
in py_action: a=2, b=4
```
I still cannot tell where the stdout from the first action ("abc efg") end, and the stdout from the second action ("in py_action: a=2, b=4") begins. This is why I'm requesting this feature.

schettino72:
If it is just for debugging you can easily:

```
'actions' : ['echo Executing cmd X', 'X',
                 'echo Executing function foo(), 'foo()']
```

`title` is for a task, so it wont work for your use-case.

I dont think messing with `verbosity` it would be a good idea, adding a new parameter would be ok.

[...]

PR #392, opened 2021-04-10
Implement CmdAction.title() and PythonAction.title()

Jasha10:
Related to #390

### DIFF
--- a/doit/action.py
+++ b/doit/action.py
@@ -7,6 +7,7 @@ from io import StringIO
 import inspect
 from pathlib import PurePath
 from threading import Thread
+from typing import Union
 import pdb
 
 from .exceptions import InvalidTask, TaskFailed, TaskError
@@ -26,6 +27,9 @@ class BaseAction(object):
 
     # must implement:
     # def execute(self, out=None, err=None)
+
+    # must implement:
+    # def title(self)
 
     @staticmethod
     def _prepare_kwargs(task, func, args, kwargs):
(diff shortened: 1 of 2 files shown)


Jasha10:
@schettino72 let me know if you have any feedback :)

schettino72:
- [ ] add new command line option --show-action (better name suggestions welcome)
- [x] implement a title() method for PythonAction & CmdAction
- [ ] Reporter will also need a new method to be called before each action

keep going :smile:

Jasha10:
>     * [ ]  Reporter will also need a new method to be called before each action

I am confused about this. The reporter is not looping over actions. The `for action in ...` loop is inside `task.py:Task.execute`:
```python
    def execute(self, stream):
        ...
        task_stdout, task_stderr = stream._get_out_err(self.verbosity)
        for action in self.actions:
            action_return = action.execute(task_stdout, task_stderr)
            ...
```

schettino72:
So we need to make Task.execute call the reporter for each `action`. *Every* output must be through a reporter.

Jasha10:
> So we need to make Task.execute call the reporter for each `action`. _Every_ output must be through a reporter.

I am planning to add a `reporter` argument to the `Task.execute` function signature, so that `Task.execute` can access the reporter.

Jasha10:
I was only able to support `ConsoleReporter`. I tried to support `JsonReporter` but there were lots of exceptions related to `task_dispatcher` that I was't able to debug.

PR #394, opened 2021-04-13
remove redundant code (overwrite_verbosity)

Jasha10:


doit/runner.py
```diff
@@ -113,7 +113,6 @@ def select_task(self, node, tasks_dict):
 
             # overwrite with effective verbosity
             task.overwrite_verbosity(self.stream)
-            task.verbosity = self.stream.effective_verbosity(task.verbosity)
```
Jasha10, line 116: These two lines of code are doing the exact same thing, so I deleted one of the lines.

PR #396, opened 2021-04-27
Allow database-less execution, considering only file_dep and target mtimes

JohannesBuchner:
Dear all,

The attached patch allows using doit in a database-less fashion. Similar to make, only the file modification times of the target and file_deps as stored in the file system are used.

The patch introduces 

* a "no-op" database (TemporaryDB), which starts empty and does not store anything to disk. It can be selected with backend: "temp".
* a "no-op" checker (NoChecker), which always returns True. This required moving a None check (task not in database) from the generic code into the other checkers.
* a uptodate function (targets_uptodate), which checks that the output files are newer than the dependent files.

This enables the way I would like to use doit. I hope it is of interest to you as well.

Below is an example dodo.py file which demonstrates the usage / configuration.

```
from doit.tools import targets_uptodate

def task_create_bar():
    return {
        'actions': ['echo creating bar', 'wc -l dodo.py > bar'],
        'targets': ['bar'],
        'file_dep': ['dodo.py'],
        'uptodate': [targets_uptodate],
        }

def task_create_foo():
    return {
        'actions': ['echo creating foo', 'cp bar foo', 'chmod 750 foo'],
        'targets': ['foo'],
        'file_dep': ['bar'],
        'uptodate': [targets_uptodate],
        }

DOIT_CONFIG = {
	'backend': 'temp',
        'check_file_uptodate': 'none',
}
```

### DIFF
--- a/doit/cmd_base.py
+++ b/doit/cmd_base.py
@@ -8,7 +8,7 @@ from .globals import Globals
 from . import version
 from .cmdparse import CmdOption, CmdParse
 from .exceptions import InvalidCommand, InvalidDodoFile
-from .dependency import CHECKERS, DbmDB, JsonDB, SqliteDB, Dependency, JSONCodec
+from .dependency import CHECKERS, DbmDB, JsonDB, TemporaryDB, SqliteDB, Dependency, JSONCodec
 from .action import CmdAction
 from .plugin import PluginDict
 from . import loader
@@ -524,7 +524,7 @@ class DoitCmdBase(Command):
 
     def get_backends(self):
         """return PluginDict of DB backends, including core and plugins"""
-        backend_map = {'dbm': DbmDB, 'json': JsonDB, 'sqlite3': SqliteDB}
+        backend_map = {'dbm': DbmDB, 'json': JsonDB, 'temp': TemporaryDB, 'sqlite3': SqliteDB}
         # add plugins
         plugins = PluginDict()
         plugins.add_plugins(self.config, 'BACKEND')
--- a/doit/cmd_base.py
+++ b/doit/cmd_base.py
@@ -8,7 +8,7 @@ from .globals import Globals
 from . import version
 from .cmdparse import CmdOption, CmdParse
 from .exceptions import InvalidCommand, InvalidDodoFile
-from .dependency import CHECKERS, DbmDB, JsonDB, TemporaryDB, SqliteDB, Dependency, JSONCodec
+from .dependency import CHECKERS, DbmDB, JsonDB, SqliteDB, Dependency, JSONCodec
 from .action import CmdAction
 from .plugin import PluginDict
 from . import loader
@@ -524,7 +524,7 @@ class DoitCmdBase(Command):
 
     def get_backends(self):
         """return PluginDict of DB backends, including core and plugins"""
-        backend_map = {'dbm': DbmDB, 'json': JsonDB, 'temp': TemporaryDB, 'sqlite3': SqliteDB}
+        backend_map = {'dbm': DbmDB, 'json': JsonDB, 'sqlite3': SqliteDB}
         # add plugins
         plugins = PluginDict()
         plugins.add_plugins(self.config, 'BACKEND')
--- a/doit/dependency.py
+++ b/doit/dependency.py
@@ -47,6 +47,55 @@ def get_file_md5(path):
     return md5.hexdigest()
 
 
+class TemporaryDB(object):
+    """Forgetful backend, in-memory only."""
+
+    def __init__(self, name, codec):
+        """Open/create a DB file"""
+        self._db = {}
+        self.name = name
+        self.codec = codec
+
+    def _load(self):
+        """load db content"""
+        pass
+
+    def dump(self):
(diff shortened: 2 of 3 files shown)


schettino72:
Could you elaborate a little more on your use-case?

I understand why someone would want to have a check based on timestamp.
But many features of doit would just not work without a DB.

Your solution is a hack (in the sense that you solved a problem even though doit was not designed to do what you want).
So I would rather have more through out changes to support you use-case then doing hacks like this.

I guess the problem is that doit has an "implicit" check on how a task is considered to be up-to-date.
`uptodate` allows you to extend it but NOT to disable the default behaviour.
Maybe having a way to disable that or replace with something like your `target_uptodate` would be preferable.
And with that in place would you still need a "no-op" TemporaryDB?

JohannesBuchner:
My use case are data analysis pipelines, which can take a long time to execute. 
I may have already run a particular command and created a output file, and want to add this as a doit recipe, in addition with a command that transforms the output file into yet another file.

At the moment, doit requires rerunning the first command, because it does not have the task in its database. Even though the output file exists and is newer than the dependencies, it insists on rerunning the task. That is a problem.

I think you placed your finger on the issue: there is some default behaviour, that currently cannot be configured/extended away. 

You are right, the "no-op" TemporaryDB is not necessary. I removed it. The updated PR is simpler.

schettino72:
Take a look at `reset-dep`. I guess it already solves your problem.

 https://pydoit.org/cmd_other.html#reset-dep

JohannesBuchner:
It works! Thanks.

FrankStain:
@schettino72 , hi there!
I can put my penny here. Recently i also had implemented the disposable database. And, yep, i think it will be useful for doit to implement the predefined disposable database to store the values only for current launch.
Some tools just can't rely on results of previous launches.

My disposable backend currently is used in workflow generators. My workspace is represented by folder structure, where the C++, Java, Python and C# code organized. Projects and solutions for all this code are generated from universal project configurations using doit.
Inside of doit, each project is represented as number of different tasks, which are configured by configuration file of project. Such generator helps me to generate the project and solution files for MS Visual Studio, XCode, PyCharm, Android Studio, ndk-build and cmake from single project configuration. But on each launch all tasks should be performed to produce the valid output.
Also, to pass the project configuration into task creator i use my own task loader. So the targets, dependencies and saved values of task with same name may differ between launches.
In this situation i need all the great functionality of doit, but only for current launch.

Currently i set the `extra_config` for `DoitMain` to set the disposable backend. But i also have one nasty doit recipe, where the code abuses JSON backend. Current JSON backend relies on `doit.globals.Globals.dep_manager._closed`, so in most first setup task is just set `Globals.dep_manager._closed = True` and no database files are created on lauch.
It will be good to set `'backend': 'disposable'` or `'backend': 'temporary'` in `DOIT_CONFIG` of such recipe instead of hacking the `Globals.dep_manager._closed` with JSON backend.

schettino72:
@JohannesBuchner `reset-dep` is an often missed feature. What makes me wonder if its name and description are good enough. have you searched the docs? which terms you have used? maybe should add an entry on FAQ...

schettino72:
@FrankStain  ok. so please open another PR or issue.

I think `temporary` is not a good a name. `disposable` is better. What about `Nonce`?

JohannesBuchner:
> @JohannesBuchner `reset-dep` is an often missed feature. What makes me wonder if its name and description are good enough. have you searched the docs? which terms you have used? maybe should add an entry on FAQ...

I was looking hard at the checker and uptodate documentation. Both would be good places to include a link, explaining that there is a way to prepare the database based on the status of the files on disk, namely reset-dep. "reset dependencies" is a strange name for the way I want to use it here, because there is no database to be reset, and it is not about the dependencies, but the status of the file system. ... "prepare database" or "recalculate/fetch file status" would be closer.

It is a bit odd to have checkers and uptodate, which have somewhat overlapping goals. That the checkers and their database are a required, prominent concept in the documentation is a bit strange to me -- having this optional and only use uptodate functions would also work well with doit, I believe. doit is quite flexible and customizable, which I appreciate. Therefore it is surprising to see quite specific code in Dependency.get_status.

I looked at the FAQ page only after I opened this PR.

JohannesBuchner:
@FrankStain 
Is your implementation similar to mine? https://github.com/pydoit/doit/pull/396/commits/d7b4618f697738f93618e0a542b9a3391867b731#diff-07223332aedc1f8f6f760f54aa4eda3be1b905a687467514d7ea10f01a23b9c0

I guess you are setting 'check_file_uptodate' to your class. That's a smart option I wasn't aware of.

schettino72:
`reset` is supposed to mean: to set the value/state a second time.

> That the checkers and their database are a required, prominent concept in the documentation is a bit strange to me

Really, I guess they are discussed only in the `run` command reference. And database later discussed as a plugin.

I think adding a note when `uptodate` is introduce is a good idea.

FrankStain:
@JohannesBuchner , yep, my implementation is kinda similar to your `TemporaryDB`.
@schettino72 `Nonce` is great! So i'll make PR soon.

PR #398, opened 2021-05-03
Require setuptools, doit/plugin.py and tests/test_plugin.py import pkg_resources

hrnciar:


schettino72:
I dont remember exactly why but some people dont like setuptools.
I guess it should be added as `extras_require` and on `dev_requirements.txt`.
And docs as well.

Potential problem is that if setuptools is missing might be hard to figure out why a plugin is not being loaded.
Maybe there should be a way to list the loaded plugins as well.

PR #404, opened 2021-09-28
Allow passing empty strings to short options + Check 'default' values

romainletendart:


romainletendart:
Tests are missing but I'm not sure which test file would be the best place to test each bug fix.
I would very much like some hints on that matter.

schettino72:
Thanks, could you give a description and example of what are the bugs you are fixing.
Looking at the changeset I have no idea what you are trying to do.

Each module has its respective unit-test module.

PR #405, opened 2021-10-15
Add field 'meta' to tasks that can be used by plugins

mlell:
The field is not interpreted by doit, but it can be used by plugins

A proposal according to the [conversation on the mailing list](https://groups.google.com/g/python-doit/c/nGzyDvD9QHc/m/qYjGc_clAAAJ). 

I am not sure about the name of the field. "meta" is a bit abstract, but it is short. Any other suggestions?
Maybe "hints"? Or "plugins"...?

Also I don't know where to put this in the docs...


schettino72:
merged. thanks.

PR #407, opened 2021-11-28
Enable file checker to handle remote storage.

samuelsinayoko:
Make FileChangedChecker more flexible (see issue https://github.com/pydoit/doit/issues/408): currently assumes file_stat is
provided by os.stat which works only for local files. Make this
customizable by adding two methods
- exists: used by Dependency class to check if target(s) exist. Defaults
to os.path.exists
- info: used by Dependency class to compute the file's stats. Defaults
to os.stat

With the addition of these two methods it's easy to handle AWS S3 and
other cloud storage.

### DIFF
--- a/doit/dependency.py
+++ b/doit/dependency.py
@@ -354,6 +354,14 @@ class SqliteDB(object):
 class FileChangedChecker(object):
     """Base checker for dependencies, must be inherited."""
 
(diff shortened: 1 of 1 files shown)


schettino72:
merged thanks.
feel free to add your name on authors file

PR #411, opened 2021-12-19
Correct behavior for default

scr-oath:
when `_CMDLINE_VARS` is unitialized, return `default`; not `None`

### DIFF
--- a/doit/doit_cmd.py
+++ b/doit/doit_cmd.py
@@ -39,7 +39,7 @@ def get_var(name, default=None):
     # This is a work-around for Windows multi-processing
     # See https://github.com/pydoit/doit/issues/164
     if _CMDLINE_VARS is None:
-        return None
+        return default
     return _CMDLINE_VARS.get(name, default)
 
 def set_var(name, value):


schettino72:
I guess you are running this on Windows, right?

Actually returning `None` was kind of intentional...
This code supposedly will be executed when you use multiprocessing, but value wont be really used.

For your code, If you actually pass a value from command line will it be used?

schettino72:
Are you using `get_var` inside an "action"?
Not supposed to be used like that, guess docs are not clear on this aspect.

scr-oath:
Hi, was running on mac, actually, within PyCharm via the recommended main boilerplate

```python
if __name__ == '__main__':
    doit.run(globals())
```

When running like this, the CLI isn't setup as it is when invoking via `doit run ...`

I really don't know any reason why returning the default when you don't have a setting would be desirous, but you say it was intentional, so please explain the rationale to me if you would.

scr-oath:
Hmm… maybe I was playing with some other condition - I can't reproduce any longer, but still - why would you ever want the default not to be returned - part of it's benefit is being able to call the function (like dict.get(key, default=...)) and not have to worry about it ever being the wrong type (like None).

schettino72:
> I really don't know any reason why returning the default when you don't have a setting would be desirous

You are right, it should never happen.
The correct line to be execute is https://github.com/pydoit/doit/blob/master/doit/doit_cmd.py#L42

The return `None` only exists to avoid error when using multiprocessing module.
But if it ever gets used, it is indication that something else is wrong.
I set it to None exactly to make it clear to identify the origin of the problem.

If you are not able to reproduce the problem I will just close this.
Let me know if you can reproduce it.

Thanks

PR #417, opened 2022-02-15
Bugfix for issue 416

m2march:


schettino72:
Thanks for report. I guess it is easier to use the "is" operator.
Can you check? https://github.com/pydoit/doit/commit/2d9e537bbb5e06cd14eec7466aa351de43d55ea5

I make a new release after you confirm this is ok.

PR #419, opened 2022-03-05
create class variable to hold config file entries prefix

castalheiro:
I intend to customize the configuration file entries prefix (e.g.tool.myapp.tasks.make_cookies instead of tool.doit.tasks.make_cookies). However, there is no global setting to replace all "doit" references with my own application name. Therefore i ended up creating a class variable that i can later modify without having to subclass DoitMain and DoitConfig classes.

### DIFF
--- a/doit/doit_cmd.py
+++ b/doit/doit_cmd.py
@@ -52,6 +52,9 @@ class DoitConfig():
     _TOML_LIBS = ['toml', 'tomlkit']
     PLUGIN_TYPES = ['command', 'loader', 'backend', 'reporter']
 
+    # configuration file entry prefix
+    PREFIX = 'tool.doit'
+
     def __init__(self):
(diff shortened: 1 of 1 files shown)


schettino72:
Hey, this is a good use-case.

Right now the code assumes `tool.doit` prefix for `pyproject.toml` and no prefix for other names of TOML files...
Since we are making the prefix configurable it would be better to make it configurable on every case (whatever TOML file name).

Are you sub-classing `DoitMain`? I would think it makes more sense to make the configuration by changes in `DoitMain` instead of `DoitConfig`.

schettino72 (changes requested):
- [ ] Configurable through DoitMain instead of DoitConfig
- [ ] Flexible prefix configuration for any TOML filename (not only pyproject.toml)

PR #423, opened 2022-04-08
support tomli and make it the default TOML provider

mgorny:
Add support for the modern `tomli` module and specify it in dependency
files instead of the obsolete `toml` library.  `tomli` is the common
choice among new Python projects and unlike `toml`, it implements
TOML 1.0.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -33,11 +33,14 @@ This is the preferred configuration source, and may gain features not available 
 
 .. note::
 
-   As a TOML parser is _not_ yet part of the standard library, a third-party package is
+   A TOML parser (`PEP 680 <https://peps.python.org/pep-0680/` / `tomllib
+   <https://docs.python.org/3.11/library/tomllib.html>`) is part of the standard
+   library since Python 3.11. For earlier Python versions, a third-party package is
    required, one of:
 
-   - `toml <https://pypi.org/project/toml/>`__
+   - `tomli <https://pypi.org/project/tomli/>`__
    - `tomlkit <https://pypi.org/project/tomlkit/>`_
+   - `toml <https://pypi.org/project/toml/>`__
 
 
 TOML vs INI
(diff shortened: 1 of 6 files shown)


doit/doit_cmd.py
```diff
@@ -49,7 +49,7 @@ def set_var(name, value):
 class DoitConfig():
     """Parse and store values taken from INI and TOML configuration files"""
     # support TOML python libs
-    _TOML_LIBS = ['toml', 'tomlkit']
+    _TOML_LIBS = ['toml', 'tomli', 'tomlkit']
```
bollwyvl, line 52: perhaps it would make sense to check `tomli` first, and move `toml` to the end, as `tomlkit` is at least still being maintained.

Further, it could start looking for [PEP 680 `tomllib`](https://peps.python.org/pep-0680/).

doc/configuration.rst
```diff
@@ -37,6 +37,7 @@ This is the preferred configuration source, and may gain features not available
    required, one of:
 
    - `toml <https://pypi.org/project/toml/>`__
+   - `tomli <https://pypi.org/project/tomli/>`__
```
bollwyvl, line 40: might also make sense to drop a link to https://peps.python.org/pep-0680/

doit/doit_cmd.py
mgorny: Sure, will do that.

doc/configuration.rst
mgorny: Yeah, rephrased the whole thing since it's in stdlib now, after all ;-).

schettino72:
@mgorny cool. thanks. 
Feel free to yourself into AUTHORS file

mgorny:
Sure, done.

schettino72:
rebased and merged. thanks.
I removed the support for the non-maintained module `toml`. 3 libs is enough!

PR #433, opened 2022-07-29
add tab completion support for gitbash

martin-eder-zeiss:
Currently tab completion does not work in gitbash, because windows does not use "/" as directory separator. The generated bash functions therefore do not have valid names. e.g.

    complete -o filenames -F _C:\Users\user\AppData\Roaming\Python\Python310\Scripts\doit C:\Users\user\AppData\Roaming\Python\Python310\Scripts\doit

### DIFF
--- a/doit/cmd_completion.py
+++ b/doit/cmd_completion.py
@@ -1,5 +1,6 @@
 """generate shell script with tab completion code for doit commands/tasks"""
 
+import os
 import sys
 from string import Template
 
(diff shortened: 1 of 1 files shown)


doit/cmd_completion.py
```diff
@@ -88,7 +89,7 @@ def _generate_bash(self, opt_values, pos_args):
             pt_list_param = ''
 
         # dict with template values
-        pt_bin_name = sys.argv[0].split('/')[-1]
+        pt_bin_name = os.path.split(sys.argv[0])[1]
```
schettino72, line 92: ok for the `os.path.split` but why have you changed `-1` to `1` ?

schettino72 (changes requested):

tom-varga:
On my way.

-Tom

doit/cmd_completion.py
martin-eder-zeiss: It doesn't really matter here. Last or second element in a [pair](https://docs.python.org/3/library/os.path.html) is the same. Do you want the -1 back?

schettino72:
merged. thanks.

PR #438, opened 2022-08-21
Correct title.py

gabalafou:
As far as I can tell, `doc/samples/title.py` is only used on the [tasks doc page](https://github.com/pydoit/doit/blob/master/doc/tasks.rst).

The **expected** output of `title.py`, as specified on the tasks page, is:

```
$ doit
.  executing... Cmd: echo abc efg
```

However, the **actual** output (before my proposed change) is:

```
doit                        
.  executing... custom_display
```

### DIFF
--- a/doc/samples/title.py
+++ b/doc/samples/title.py
@@ -1,6 +1,6 @@
 
 def show_cmd(task):
-    return "executing... %s" % task.name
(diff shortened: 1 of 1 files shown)


schettino72:
merged. thanks.

PR #466, opened 2024-05-29
Added __repr__ to the config_changes and result_dep classes.

abingham:
I added some basic `__repr__` implementations to the `config_changed` and `result_dep` classes. I needed a nicer representation for these so that I could better understand the output of the `info` command. 

If these particular implementations aren't acceptable for some reason, let me know what you'd prefer.


schettino72:
thanks

PR #489, opened 2026-09-19
Interactive TUI to explore task statii

superhellth:
I was missing a good way to find out why which tasks would rerun in my rather large doit project (35 tasks). So, I built the doit status command. 

doit status <task>: Shows whether this task is up-to-date/run/may-rerun (at least one parent task is stale), its parent- and child-tasks
doit status -i [task]: Opens an interactive TUI to explore why a certain parent/child might be stale, also displaying parents and childs

Both commands share one consistent UI.

The following screenshot shows the interactive TUI:
image

Everything is tested and i iterated properly to make this (in my mind at least) a well-polished feature. Tested the TUI myself on Linux (Ubuntu on WSL) and Windows 11.

No dependencies were added (new TUI built in doit/status_tui + doit/status_term).
All tests pass.
Docs updated accordingly.


schettino72:
Cool. I have not tried it yet but i always wanted to have a TUI for doit.

But ideally this would be a separate command and package. doit allow extension/plugins for custom sub-commands. have you looked into that ? Please try it that way, if there is anything missing to allow this kind of functionality we can improve the plugin implementation.

Then you can create a package on PYPI and update the TUI independent of doit itself. I would be glad to link the TUI project from doit official docs. what do you think ?

superhellth:
> Cool. I have not tried it yet but i always wanted to have a TUI for doit.
> 
> But ideally this would be a separate command and package. doit allow extension/plugins for custom sub-commands. have you looked into that ? Please try it that way, if there is anything missing to allow this kind of functionality we can improve the plugin implementation.
> 
> Then you can create a package on PYPI and update the TUI independent of doit itself. I would be glad to link the TUI project from doit official docs. what do you think ?

Sounds good, i will look into that. Do you think the static doit status <task> fits the main repo or is that too close to doit list --status? Also is it fine if i introduce the release() and reopen() functions to dependency.py? Without that the Plugin might get really hacky.

